PR #88, opened 2023-09-11
feat: implement chainsync ouroboros server

scarmuega:


src/serve/ouroboros/mod.rs
```diff
@@ -105,3 +83,140 @@ async fn handle_blockfetch(db: RollDB, mut peer: PeerServer) -> Result<(), Error
...
+                        let content = HeaderContent {
+                            variant: 1, // TODO
+                            byron_prefix: None,
+                            cbor: block.header().cbor().to_vec(),
+                        };
+
+                        protocol
+                            .send_roll_forward(content, tip)
+                            .await
+                            .map_err(Error::server)?;
```
jmhrpr: The first response to a RequestNext message the Haskell node makes after a successful intersect is a MsgRollback to the intersection point, so we probably want to copy that behaviour

```
    let known_points = vec![Point::Specific(
        43847831u64,
        hex::decode("15b9eeee849dd6386d3770b0745e0450190f7560e5159b1b3ab13b14b2684a45").unwrap(),
    )];

    let (point, _) = peer.chainsync().find_intersect(known_points).await.unwrap();

    info!("intersected point is {:?}", point);

    for _ in 0..3 {
        let next = peer.chainsync().request_next().await.unwrap();

        match next {
            chainsync::NextResponse::RollForward(h, _) => {
                info!("rolling forward, header size: {}", h.cbor.len())
            }
            chainsync::NextResponse::RollBackward(x, _) => info!("rollback to {:?}", x),
            chainsync::NextResponse::Await => info!("tip of chaing reached"),
        };
    }
```

```
2023-09-12T10:11:39.335692Z  INFO n2n_miniprotocols: intersected point is Some((43847831, 15b9eeee849dd6386d3770b0745e0450190f7560e5159b1b3ab13b14b2684a45))
2023-09-12T10:11:39.625239Z  INFO n2n_miniprotocols: rollback to (43847831, 15b9eeee849dd6386d3770b0745e0450190f7560e5159b1b3ab13b14b2684a45)
2023-09-12T10:11:39.958946Z  INFO n2n_miniprotocols: rolling forward, header size: 1006
2023-09-12T10:11:40.249326Z  INFO n2n_miniprotocols: rolling forward, header size: 1006
```

src/serve/ouroboros/mod.rs
```diff
@@ -105,3 +83,140 @@ async fn handle_blockfetch(db: RollDB, mut peer: PeerServer) -> Result<(), Error
...
+                    } else {
+                        protocol.send_await_reply().await.map_err(Error::server)?;
+                    }
+                }
+            },
+            None => todo!(),
```
jmhrpr: `None` here signals a ClientDone from the client, so we can probably exit with a log message saying as much and then `return Ok(())` maybe?

PR #128, opened 2023-12-06
Ouroboros ChainSync Continued

jmhrpr:


src/sync/pull.rs
```diff
@@ -205,7 +205,8 @@ impl gasket::framework::Worker<Stage> for Worker {
     }
 
     async fn teardown(&mut self) -> Result<(), WorkerError> {
-        self.peer_session.abort();
+        // TODO: doesn't work because teardown takes mutable ref but abort takes
+        // self.peer_session.abort();
```
jmhrpr, line 209: Not sure what to do here

src/serve/ouroboros/tests.rs
```diff
@@ -86,8 +99,15 @@ async fn test_blockfetch() {
     server.abort();
 }
 
-//#[tokio::test]
-async fn do_chainsync() {
+#[ignore = "broken"]
+#[tokio::test]
+async fn test_chainsync() {
```
jmhrpr, line 103: Don't know why this test won't work. It intersects, client requests the block then tries to receive the block, and the server eventually responds with a block, but the client does not see it for some reason and is stuck trying to receive it.

PR #154, opened 2024-01-28
feat(utxo-rpc): Add support for BlockFetch by Slot or Height and DumpHistory to support BlockHeight

Mercurial:
Work In Progress to support block fetch not only by hash but also by Slot or Height.

As of right now only fetching by block hash is supported.

```rs
async fn fetch_block(
        &self,
        request: Request<FetchBlockRequest>,
    ) -> Result<Response<FetchBlockResponse>, Status> {
        let message = request.into_inner();

        let blocks: Result<Vec<_>, _> = message
            .r#ref
            .iter()
            .map(|r| bytes_to_hash(&r.hash))
            .map(|hash| self.chain.get_block(hash))
            .collect();

        let out: Vec<_> = blocks
            .map_err(|_err| Status::internal("can't query block"))?
            .iter()
            .flatten()
            .map(|b| raw_to_anychain(b))
            .collect();

        let response = FetchBlockResponse { block: out };

        Ok(Response::new(response))
    }
```

Aswell as `next_token` in the `DumpHistoryResponse` needs to support Block->Height according to the `BlockRef` structure in the U5C specification.

image

I'm thinking this would need some kind of new KeyValueStore within `pallas-rolldb` similiar to

```rs
// slot => block hash
pub struct HashBySlotKV;

impl KVTable<DBInt, DBHash> for HashBySlotKV {
    const CF_NAME: &'static str = "HashBySlotKV";
}
```

it could be something like

```rs
// height => block hash
pub struct HashBySlotKV;

impl KVTable<DBInt, DBHash> for HashByHeightKV {
    const CF_NAME: &'static str = "HashByHeightKV";
}
```

but I think the main hurdle is that the block height / number doesn't seem to be a first-class-citizen field within the pipeline.

image

image

image

Looking for feedback / suggestions

### DIFF
--- a/src/serve/grpc/sync.rs
+++ b/src/serve/grpc/sync.rs
@@ -88,7 +88,7 @@ impl chain_sync_service_server::ChainSyncService for ChainSyncServiceImpl {
         request: Request<DumpHistoryRequest>,
     ) -> Result<Response<DumpHistoryResponse>, Status> {
         let msg = request.into_inner();
-        let from = msg.start_token.map(|r| r.index).unwrap_or_default();
+        let from = msg.start_token.map(|r| r.slot).unwrap_or_default();
         let len = msg.max_items as usize + 1;
 
         let mut page: Vec<_> = self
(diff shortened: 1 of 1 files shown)


Mercurial:
nvm as per discussion with Santi `BlockRef` for UTxORPC should only have `hash` and `index` to be utxo chain agnostic. So this feature is not needed

PR #158, opened 2024-02-12
Updated pallas, utxorpc sdk to support utxorpc-spec v0.3.0

Mercurial:
This PR updates dolos to support the new `utxorpc-spec 0.3.0`. 

This uses utxorpc `rust-sdk` temporary branch https://github.com/utxorpc/rust-sdk/pull/2 `Mercurial:feat/update-dependency-spec-0.3.0` 

Most likely should wait for PR to be merged and point to the merged version in `Cargo.toml` instead or a published crate version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ pallas = { git = "https://github.com/txpipe/pallas.git", features = ["unstable"]
 gasket = { version = "^0.5", features = ["derive"] }
 # gasket = { path = "../../construkts/gasket-rs/gasket", features = ["derive"] }
 
-utxorpc = { version = "1.0.0-alpha.1" }
+utxorpc = { git = "https://github.com/Mercurial/rust-sdk.git",branch = "feat/update-dependency-spec-0.3.0" }
 # utxorpc = { path = "../../utxorpc/rust-sdk" }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -18,7 +18,7 @@ pallas = { git = "https://github.com/txpipe/pallas.git", features = ["unstable"]
 gasket = { version = "^0.5", features = ["derive"] }
 # gasket = { path = "../../construkts/gasket-rs/gasket", features = ["derive"] }
 
-utxorpc = { version = "1.0.0-alpha.1" }
+utxorpc = { git = "https://github.com/Mercurial/rust-sdk.git",branch = "feat/update-dependency-spec-0.3.0" }
```
scarmuega: until the SDK has anything interesting, I would just depend on `utxorpc-spec = "0.3.0"` and avoid the extra step.

In the near future, when the SDK has some actual benefits other than a re-export of structures, we'll switch back.

Mercurial: that makes sense, will do the needed change!

scarmuega:
sorry, I missed this PR and applied the same changes in a different one.

Closing in favor of #159

PR #192, opened 2024-04-08
feat: introduce query db

scarmuega:


src/querydb/prelude.rs
```diff
@@ -0,0 +1,71 @@
...
+    TableError(redb::TableError),
+    TransactionError(redb::TransactionError),
+}
+
+pub enum ReadError {
+    ChainTipNotFound,
+    TransactionError(TransactionError),
+    TableError(TableError),
+    StorageError(StorageError),
+}
```
scarmuega: There's no need for so many errors. In can be merged into a single enum. All of the redb-related errors can be a single variant of the enum.

src/querydb/store.rs
```diff
@@ -0,0 +1,220 @@
...
+        write_tx: &'a WriteTransaction,
+        block_cbor: &'a [u8],
+    ) -> Result<MultiEraBlock<'a>, StoreError> {
+        let block: MultiEraBlock =
+            MultiEraBlock::decode(block_cbor).map_err(StoreError::BlockDecoding)?;
+        let mut block_table: Table<BlockKeyType, BlockValueType> = write_tx
+            .open_table(BLOCK_TABLE)
+            .map_err(|e| StoreError::ReDBError(ReDBError::TableError(e)))?;
+        let _ = block_table.insert(block.hash().deref(), block_cbor);
+        let mut chain_tip_table: Table<ChainTipKeyType, ChainTipValueType> = write_tx
```
scarmuega: Querying the chain tip should be about looking for the latest block, there's no need to have a custom table.

src/querydb/store.rs
```diff
@@ -0,0 +1,220 @@
...
+            .map_err(|e| StoreError::ReDBError(ReDBError::TableError(e)))?;
+        let mut utxo_by_addr_table: MultimapTable<UTxOByAddrKeyType, UTxOByAddrValueType> =
+            write_tx
+                .open_multimap_table(UTXO_BY_ADDR_TABLE)
+                .map_err(|e| StoreError::ReDBError(ReDBError::TableError(e)))?;
+        let mut utxo_by_beacon_table: MultimapTable<UTxOByBeaconKeyType, UTxOByBeaconValueType> =
+            write_tx
+                .open_multimap_table(UTXO_BY_BEACON_TABLE)
+                .map_err(|e| StoreError::ReDBError(ReDBError::TableError(e)))?;
+        for (index, output) in tx.outputs().iter().enumerate() {
```
scarmuega: Instead of using `tx.outputs()` we should use `tx.produces()` since it takes into account valid vs invalid txs.

src/querydb/store.rs
```diff
@@ -0,0 +1,220 @@
...
+                .non_ada_assets()
+                .iter()
+                .map(MultiEraPolicyAssets::policy)
+            {
+                utxo_by_beacon_table
+                    .insert(policy.deref(), (tx.hash().deref().as_slice(), index as u8))
+                    .map_err(|e| StoreError::ReDBError(ReDBError::InsertError(e)))?;
+            }
+        }
+        Ok(())
```
scarmuega, line 138: We're missing the update of the data for the consumed utxos. Eg: when an utxo is consumed in a tx, we should remove it from the tables and indices.

src/querydb/prelude.rs
MaicoLeberle: Sure! There's no unified error type in ReDB though, but I'll box all these errors dynamically

src/querydb/store.rs
MaicoLeberle: I tried doing that in the first place, but the latest block doesn't seem to be so easily retrievable. The [last](https://docs.rs/redb/latest/redb/trait.ReadableTable.html#) method for ReadableTable's returns the key-value pair corresponding to the largest key, and blocks are indexed by their hash (which don't necessarily preserve the order in which they were stored).

src/querydb/store.rs
scarmuega: @MaicoLeberle when processing consumed utxos, we should also remove the corresponding entries from the indices (eg: remove the utxo (hash, index) from the `utxo_by_address`, `utxo_by_beacon`, etc.

you can do this by retrieving the utxo body from the table and re-computing the address, tokens, etc.

src/querydb/store.rs
```diff
@@ -0,0 +1,250 @@
...
+        let block: MultiEraBlock =
+            MultiEraBlock::decode(block_cbor).map_err(StoreError::BlockDecoding)?;
+        let mut block_table: Table<BlockKeyType, BlockValueType> = write_tx
+            .open_table(BLOCK_TABLE)
+            .map_err(|e| StoreError::ReDBError(Box::new(e)))?;
+        let _ = block_table.insert(block.hash().deref(), block_cbor);
+        let mut blockchain_tip_table: Table<ChainTipKeyType, ChainTipValueType> = write_tx
+            .open_table(CHAIN_TIP_TABLE)
+            .map_err(|e| StoreError::ReDBError(Box::new(e)))?;
+        let _ = blockchain_tip_table.insert((), block.hash().deref());
```
scarmuega: we don't need a `CHAIN_TIP_TABLE`. There's nothing we need to store regarding the chain tip. We can just query the last entry from the `BLOCK_TABLE`.

MaicoLeberle: The problem is blocks are indexed by hash in `BLOCK_TABLE`. But the order between block hashes doesn't coincide with the order in which blocks are inserted through time. Still, tables in ReDB are ordered by key, so they're ordered by hashes and not by insertion order.

The current approach may be a bit unattractive from a design standpoint, but shouldn't spend too much extra memory or CPU anyways...

Issue #180, opened 2024-03-26
test: implement PParams blackbox test using Haskell node as test oracle


scarmuega:

Quantumplation:
:wave: I'm going to tackle this, if you want to assign it to me

PR #217, opened 2024-05-10
Initial pass at pparams testing

Quantumplation:
This is a first pass at testing the implementation of fold_pparams, as described in #180 

In particular, it only checks that the resulting fold results in the correct major protocol version, until we add serialization/deserialization to the MultiEraProtocolParams object (which is what I'll work on next). After we have that serialization, I'll drop the different `protocol-parameters.json` files into `src/ledger/pparams/test_data/mainnet/expected_params` and update the tests.

This tries to integrate / document the information / reality of how protocol versions have historically been treated; in particular, there is a `protocol version` that is used to negotiate in the networking stack, and there's a `protocol version` (maybe better called a `ledger version`) that is used to determine the semantics of the ledger, and they used to not agree.

I'm still not sure my comments are correct, or that I have a perfect grasp on this, but I plan to sync with @JaredCorduan on this when he has some spare cycles.

Some other things that might be useful to do:
 - Add a small local utility to test_data for "refreshing" this test data?
   - These kinds of black-box tests will be quite common, so having a pattern for being able to refresh the test data with examples from the new era would be helpful
 - Move the pparams handling into pallas itself
   - Probably best for me to leave this to @scarmuega, as he has a better vision for how everything is going to fit together


scarmuega (approved):

scarmuega:
amazing!

yes, I can move it to pallas after the next release.

PR #304, opened 2024-07-17
chore: Implement ReadParams

gonzalezzfelipe:
Depends on: https://github.com/utxorpc/spec/pull/116


scarmuega:
@gonzalezzfelipe looks very good! The only thing I'm concerned is the hack to handle the Shelley params update on preview, I'm worried of introducing tech-debt.

I think that the correct approach would be to understand which era should be used to initialize the pparams by looking at the genesis files. Not sure what, but there should be something in the configs that flags Shelley as the initial protocol version for preview.

gonzalezzfelipe:
> @gonzalezzfelipe looks very good! The only thing I'm concerned is the hack to handle the Shelley params update on preview, I'm worried of introducing tech-debt.
> 
> I think that the correct approach would be to understand which era should be used to initialize the pparams by looking at the genesis files. Not sure what, but there should be something in the configs that flags Shelley as the initial protocol version for preview.

I tried another aproach, bootstrapping the Shelley params when the initial update doesn't correspond to a Byron era update instead of going with the genesis files because I couldn't find a way to deduce it from there (other than using the network magic and hardcoding that Preview should begin from Shelley)

scarmuega (approved):

PR #326, opened 2024-08-20
fix: Use skip if not empty param

gonzalezzfelipe:


src/bin/dolos/bootstrap.rs
```diff
@@ -206,6 +206,11 @@ fn import_hardano_into_wal(
 pub fn run(config: &super::Config, args: &Args, feedback: &Feedback) -> miette::Result<()> {
     //crate::common::setup_tracing(&config.logging)?;
 
+    if args.skip_if_not_empty & crate::common::open_wal(config).is_ok() {
+        info!("Skipping bootstrap, data already present.");
+        return Ok(());
+    }
+
```
scarmuega: I think this premature `open_wal` could lock the second one that does the actual import if the struct is not dropped.

gonzalezzfelipe: oh, didn't know that. I'll check, and maybe put this in a closure to make sure it is dropped

gonzalezzfelipe:
This is already the behaviour now, so closing the PR

PR #332, opened 2024-09-04
chore: add num-integer-backend feature to mithril dependency to fix build issue

Mercurial:


scarmuega:
@Mercurial this change has a penalty on performance.

Could we go back and try to understand the root cause of the issue? because I can't reproduce the build error you're describing.

PR #400, opened 2024-12-16
Inital `opentelemetry` layout

### LINKED ISSUE #399: OpenTelemetry logging
Right now the logs are just printed to console as per https://dolos.txpipe.io/running/logging

For Paima, we need to aggregate the logs of many services across many chains. It's much easier for us if every part of the stack integrates opentelemtry so that we can aggregate all the logs in a unified format (instead of everybody having their own opinionated log formatting)

SebastienGllmt:
I added code to get opentelemtry working so I could consume it from Paima Engine

Closes #399

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ serde_json = "1.0.122"
 thiserror = "1.0.30"
 lazy_static = "1.4.0"
 tracing = "0.1.37"
-tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
+tracing-subscriber = "0.3.17"
 opentelemetry-appender-tracing = { version = "0.27.0", default-features = false}
 opentelemetry = { version = "0.27.0", default-features = false, features = ["trace"] }
 opentelemetry_sdk = { version = "0.27.0", default-features = false, features = ["trace", "rt-tokio"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,12 @@ serde_json = "1.0.122"
 thiserror = "1.0.30"
 lazy_static = "1.4.0"
 tracing = "0.1.37"
-tracing-subscriber = "0.3.17"
+tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
+opentelemetry-appender-tracing = { version = "0.27.0", default-features = false}
+opentelemetry = { version = "0.27.0", default-features = false, features = ["trace"] }
+opentelemetry_sdk = { version = "0.27.0", default-features = false, features = ["trace", "rt-tokio"] }
+opentelemetry-http = { version = "0.27.0", default-features = false}
+opentelemetry-otlp = { version = "0.27.0", features = ["logs", "http-json", "reqwest-client"] }
 bincode = "1.3.3"
 miette = { version = "7.2.0", features = ["fancy"] }
 tokio = { version = "^1.40", features = ["rt", "rt-multi-thread", "signal"] }
--- a/src/bin/dolos/common.rs
+++ b/src/bin/dolos/common.rs
@@ -75,10 +75,9 @@ pub fn setup_tracing(config: &LoggingConfig) -> miette::Result<()> {
         .with_protocol(Protocol::HttpJson)
         .build().unwrap();
 
-    // Create a new OpenTelemetry trace pipeline that prints to stdout
     let provider = LoggerProvider::builder()
-        // .with_simple_exporter(exporter)
-        .with_batch_exporter(exporter, runtime::Tokio)
+        .with_simple_exporter(exporter)
+        // .with_batch_exporter(exporter, runtime::Tokio)
         .build();
     let layer = layer::OpenTelemetryTracingBridge::new(&provider);
 
--- a/src/bin/dolos/common.rs
+++ b/src/bin/dolos/common.rs
@@ -9,6 +9,13 @@ use tokio::task::JoinHandle;
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, warn};
 use tracing_subscriber::{filter::Targets, prelude::*};
+use opentelemetry_appender_tracing::layer;
+use opentelemetry_otlp::WithExportConfig;
+use opentelemetry_otlp::{LogExporter, Protocol};
+use opentelemetry_sdk::{
+    logs::LoggerProvider,
+    runtime,
+};
 
 use dolos::prelude::*;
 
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -32,6 +32,11 @@ thiserror = "1.0.30"
 lazy_static = "1.4.0"
 tracing = "0.1.37"
 tracing-subscriber = "0.3.17"
+opentelemetry-appender-tracing = { version = "0.27.0", default-features = false}
+opentelemetry = { version = "0.27.0", default-features = false, features = ["trace"] }
+opentelemetry_sdk = { version = "0.27.0", default-features = false, features = ["trace", "rt-tokio"] }
+opentelemetry-http = { version = "0.27.0", default-features = false}
+opentelemetry-otlp = { version = "0.27.0", features = ["logs", "http-json", "reqwest-client"] }
```
SebastienGllmt, line 39: note: I used the `http+json` version of opentelemetry instead of the grpc+protobuf

I did this because we currently only support the `http+json` approach in Paima at the moment, but plan to support grpc+protobuf in the future

src/bin/dolos/common.rs
```diff
@@ -62,6 +69,19 @@ pub fn open_data_stores(config: &crate::Config) -> Result<Stores, Error> {
 pub fn setup_tracing(config: &LoggingConfig) -> miette::Result<()> {
     let level = config.max_level;
 
+    let exporter = LogExporter::builder()
+        .with_http()
+        .with_endpoint("http://localhost:4318/v1/logs")
```
SebastienGllmt, line 74: note: this should be configured automatically for you from an env variable (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp), but I just set it explicitly here

src/bin/dolos/common.rs
```diff
@@ -62,6 +69,19 @@ pub fn open_data_stores(config: &crate::Config) -> Result<Stores, Error> {
...
+        .with_protocol(Protocol::HttpJson)
+        .build().unwrap();
+
+    // Create a new OpenTelemetry trace pipeline that prints to stdout
+    let provider = LoggerProvider::builder()
+        // .with_simple_exporter(exporter)
```
SebastienGllmt, line 80: using `with_simple_exporter` is usually better for dev environments (it's significantly more responsive)

src/bin/dolos/main.rs
```diff
@@ -186,7 +186,8 @@ impl Config {
     }
 }
 
-fn main() -> Result<()> {
+#[tokio::main]
+async fn main() -> Result<()> {
```
SebastienGllmt, line 190: If I don't have this, `cargo run sync` gives me the error
```
thread 'main' panicked at /home/sebdev20/.cargo/registry/src/index.crates.io-6f17d22bba15001f/hyper-util-0.1.5/src/client/legacy/connect/dns.rs:121:24:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Note that I don't need this to run `cargo run daemon` - that one instead starts but then deadlocks right away. I tried switching to `with_batch_exporter` which doesn't deadlock, but instead of doesn't seem to actually send anything to opentelemetry either (not sure why)

src/bin/dolos/common.rs
```diff
@@ -62,6 +69,18 @@ pub fn open_data_stores(config: &crate::Config) -> Result<Stores, Error> {
...
     let level = config.max_level;
 
+    let exporter = LogExporter::builder()
+        .with_http()
+        .with_endpoint("http://localhost:4318/v1/logs")
+        .with_protocol(Protocol::HttpJson)
+        .build().unwrap();
+
+    let provider = LoggerProvider::builder()
+        .with_simple_exporter(exporter)
```
SebastienGllmt, line 79: we want to use `with_simple_exporter` since it's a lot more responsive in dev mode (and for Paima you're running everything on the same machine so the network overhead is minimal), but for production where you have different machines `with_batch_exporter` will work a lot better

src/bin/dolos/common.rs
```diff
@@ -9,6 +9,13 @@ use tokio::task::JoinHandle;
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, warn};
 use tracing_subscriber::{filter::Targets, prelude::*};
+use opentelemetry_appender_tracing::layer;
+use opentelemetry_otlp::WithExportConfig;
+use opentelemetry_otlp::{LogExporter, Protocol};
```
SebastienGllmt, line 14: Note: I only export logs and not traces (spans) or metrics just because I only needed logs to make sure it works

Cargo.toml
```diff
@@ -32,6 +32,11 @@ thiserror = "1.0.30"
 lazy_static = "1.4.0"
 tracing = "0.1.37"
 tracing-subscriber = "0.3.17"
+opentelemetry-appender-tracing = { version = "0.27.0", default-features = false}
```
SebastienGllmt, line 35: note: I only added the system to forward the `tracing` crate to opentelemtry, but not other crates

that is to say, I noticed in Dolos (and maybe downstream crates?) you occasionally use `use log::` which won't be forwarded

src/bin/dolos/main.rs
SebastienGllmt: I realize now that unfortunately this breaks `cargo run daemon` with the following error

```
thread 'main' panicked at /home/me/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.40.0/src/runtime/scheduler/multi_thread/mod.rs:86:9:
Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
```

note: `cargo run sync` still works

SebastienGllmt: They recently (December 2024) updated the docs on how to get the `SimpleExporter` working to avoid the deadlock mentioned earlier

However, it turns out that, unfortunately, using the `SimpleExporter` requires a fairly different setup than the batch exporter:

> `SimpleExporter` requires enabling feature flag `reqwest-blocking-client` and making the `main()` a normal main and *not* `tokio::main`

However, I can't seems to get it to work because
- `reqwest-blocking-client` as recommended errors with https://github.com/seanmonstar/reqwest/issues/1017
- using `reqwest-client` gives `there is no reactor running, must be called from the context of a Tokio 1.x runtime` (and adding this runs into the issue with `cargo run daemon` mentioned above)

There are some issues that were resolved in opentelemetry that *sound* like they might be relevant such as https://github.com/open-telemetry/opentelemetry-rust/pull/2431, but they are not released yet. More generally, I wonder if https://github.com/open-telemetry/opentelemetry-rust/issues/2386 might make the integration easier

PR #409, opened 2024-12-22
Add max_slots_before_prune config

MartinSchere:
This PR adds a new config called `max_slots_before_prune` which allows the user to configure the oldest age of the slot kept in the ledger.

This is useful if the data in the ledger is required for use cases other than syncing the chain, such as indexers.

### DIFF
--- a/src/state/mod.rs
+++ b/src/state/mod.rs
@@ -220,6 +220,7 @@ pub fn apply_block_batch<'a>(
     store: &mut LedgerStore,
     byron: &byron::GenesisFile,
     shelley: &shelley::GenesisFile,
+    max_slots_before_prune: Option<u64>,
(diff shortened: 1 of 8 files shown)


src/state/mod.rs
```diff
@@ -238,7 +239,10 @@ pub fn apply_block_batch<'a>(
         .map(|x| x.0)
         .unwrap();
 
-    let to_finalize = lastest_immutable_slot(tip, byron, shelley);
+    let to_finalize = max_slots_before_prune
+        .map(|x| tip - x)
```
MartinSchere, line 243: @scarmuega I'm not sure if you want extra guardrails here in case the user shoots themselves in the foot by setting `max_slots_before_prune` to a very small number.

PR #422, opened 2024-12-30
fix: final protocol param update not being applied

ricomiles:
This fixes a bug where the last two protocol parameter updates aren't being applied. The issue arises because the epoch for-loop does not include the last two updates due to how the filtering of the updates works.
This also fixes protocol version param not updating


scarmuega:
@ricomiles are we sure this is the correct behavior?
if this is the fix, then the problem should only appear when you're evaluating the updates in the exact same epoch as when they were submitted. Is that what's happening?

PR #424, opened 2024-12-30
feat: allow tx eval to return original cbor

ricomiles:
This adds the original_cbor field to TxEvalResult and also includes original_cbor in tx_eval_to_u5c function.


scarmuega:
@Mercurial what was the use-case for having the original cbor in the payload?
I know we discussed it but I can't remember. It does seems redundant if you're sending it in the eval request.

PR #428, opened 2025-01-13
Create MainBranch Ci 

leo42:
Create dockerCi.yml

This Ci runs on every commit to main and creates a docker image with the tag "main"


scarmuega:
@leo42 I would prefer if we modify the existing docker workflow instead of adding a new one. I think that to accomplish what you want we just need to adjust the trigger so that it runs on every push to main.

leo42:
@scarmuega  I merged it into a single file with filters for the tags to be issued only when you cut a release

.github/workflows/docker.yml
```diff
@@ -74,9 +76,10 @@ jobs:
           images: ghcr.io/txpipe/dolos
           tags: |
             type=raw,value=latest,enable={{is_default_branch}}
-            type=semver,pattern=v{{major}}
-            type=semver,pattern=v{{major}}.{{minor}}
-            type=semver,pattern=v{{version}}
+            type=raw,value=stable,enable=${{ startsWith(github.ref, 'refs/tags/v') }}
+            type=semver,pattern=v{{major}},enable=${{ startsWith(github.ref, 'refs/tags/v') }}
+            type=semver,pattern=v{{major}}.{{minor}},enable=${{ startsWith(github.ref, 'refs/tags/v') }}
+            type=semver,pattern=v{{version}},enable=${{ startsWith(github.ref, 'refs/tags/v') }}
```
scarmuega, line 82: Are you sure we need for an explicit `enable` for semver tags?
I was under the impression that if the git tag is not present, the action will just skip semver.

leo42: Unsure, but better safe than sorry

scarmuega (approved):

PR #447, opened 2025-02-05
chore: Add endpoints to minibf

gonzalezzfelipe:


src/serve/minibf/mod.rs
```diff
@@ -16,7 +16,11 @@ pub struct Config {
 
 pub async fn serve(
     cfg: Config,
+    genesis: Arc<Genesis>,
+    genesis_values: GenesisValues,
```
scarmuega: GenesisValues from traverse is deprecated. We shouldn't need this. Why is it used for?

Cargo.toml
```diff
@@ -12,8 +12,8 @@ authors = ["Santiago Carmuega <santiago@carmuega.me>"]
 
 
 [dependencies]
-pallas = { git = "https://github.com/txpipe/pallas.git", features = ["hardano", "applying"] }
-# pallas = { version = "^0.30.1", features = ["hardano", "applying"] }
+pallas = { git = "https://github.com/gonzalezzfelipe/pallas.git", branch = "fix/error-on-tx-validation", features = ["hardano", "applying"] }
```
scarmuega: this fix is already merged

src/serve/minibf/mod.rs
gonzalezzfelipe: This ones: https://github.com/txpipe/pallas/blob/c08d766d38604d13786863332a99edb99455189b/pallas-traverse/src/time.rs#L121-L144

scarmuega (approved):

PR #451, opened 2025-02-19
feat: Chain

gonzalezzfelipe:
missing:

* [x] Replace wall on all queries
* [x] Change prune_history from WAL to chain, keeping WAL as mutable part only.
* [x] Add Chain indexes:
    * [x] BlockHash
    * [x] BlockNumber
    * [x] Address
    * [x] AddressPaymentPart
    * [x] AddressDelegationPart
    * [x] Policy
    * [x] PolicyAsset
    * [x] Transactions
    * [x] ScriptHash (AuxData on txs)
    * [x] DatumHash (AuxData on txs)
 * [x] Improve init procedure to handle history pruning better


src/model.rs
```diff
@@ -53,11 +53,17 @@ pub struct StorageConfig {
...
+    pub chain_cache: Option<usize>,
+
+    /// Whether to remove inmutable part from WAL. Defaults to true
+    pub prune_wal: Option<bool>,
 
     /// Maximum number of slots to keep in the ledger before pruning
     pub max_ledger_history: Option<u64>,
+
+    /// Maximum number of slots (not blocks) to keep in Chain
+    pub max_chain_history: Option<u64>,
```
gonzalezzfelipe: @scarmuega these are the relevant changes in config. I was thinking of the following for backwards compatibility.

* Derive `prune_wal` from `max_wal_history`. If `max_wal_history` is None, then `prune_wal`  will be treated as `false` to avoid erasing WAL for user that might want to keep it.
* Derive `max_chain_history` from `max_wal_history`. If `max_wal_history` is defined and `max_chain_history` is not, then `max_chain_history` is taken from it.

The only "problem" that this has is that for the default chase, where `max_wal_history` is None, we will be populating the `chain` and the WAL, and nothing will be pruned, but I don't see it as a major issue.

Other than that, I don't see major issues. Chain will be populated since version update, and if someone wants to add the historic they should bootstrap again (I find that reasonable)

src/bin/dolos/init/include/mod.rs
```diff
@@ -36,3 +37,31 @@ pub fn save_genesis_configs(root: &Path, network: &super::KnownNetwork) -> miett
...
+        super::KnownNetwork::CardanoPreview => Genesis {
+            alonzo: serde_json::from_slice(preview::ALONZO).unwrap(),
+            conway: serde_json::from_slice(preview::CONWAY).unwrap(),
+            byron: serde_json::from_slice(preview::BYRON).unwrap(),
+            shelley: serde_json::from_slice(preview::SHELLEY).unwrap(),
+            force_protocol: Some(6),
+        },
+    };
+    ((3.0 * genesis.byron.protocol_consts.k as f32) / (genesis.shelley.active_slots_coeff.unwrap()))
+        as u64
```
scarmuega: this is a ledger concern. We should have a function in ledger to compute the mutability window from a genesis object. Something [like this](https://github.com/txpipe/dolos/blob/ae1ade251a4ac446b81e8ad0d3074d343101696e/src/ledger/mod.rs#L251)

src/chain/redb/indexes/address_payment_part.rs
```diff
@@ -0,0 +1,31 @@
...
+use std::hash::{DefaultHasher, Hash as _, Hasher};
+
+use crate::model::BlockSlot;
+
+type Error = crate::chain::ChainError;
+
+pub struct AddressPaymentPartApproxIndexTable;
+impl AddressPaymentPartApproxIndexTable {
+    pub const DEF: MultimapTableDefinition<'static, u64, u64> =
+        MultimapTableDefinition::new("addresspaymentpartapproxindex");
```
scarmuega: I would revisit the naming convention:
- `aproxindex` suffix seems irrelevant
- `part` suffix seems irrelevant
- `address` can be inferred from `payment`
- add `by` prefix as convention to identify indexes (what we use in state store)

This would end up being `bypayment`. Happy to discuss other alternatives, but I would keep it as short as possible.

src/bin/dolos/doctor/rebuild.rs
```diff
@@ -54,24 +60,15 @@ pub fn run(config: &crate::Config, _args: &Args, feedback: &Feedback) -> miette:
...
-        .into_diagnostic()
-        .context("finding ledger cursor")?
-        .map(|ledger::ChainPoint(s, h)| wal.assert_point(&wal::ChainPoint::Specific(s, h)))
-        .transpose()
-        .into_diagnostic()
-        .context("locating wal sequence")?;
-
     // Amount of slots until unmutability is guaranteed.
     let lookahead = ((3.0 * genesis.byron.protocol_consts.k as f32)
         / (genesis.shelley.active_slots_coeff.unwrap())) as u64;
```
scarmuega: this value should come from a function in the ledger.

src/bin/dolos/doctor/mod.rs
```diff
@@ -2,13 +2,13 @@ use clap::{Parser, Subcommand};
...
-mod rebuild_ledger;
+mod rebuild;
 mod wal_integrity;
 
 #[derive(Debug, Subcommand)]
 pub enum Command {
-    /// rebuilds the whole ledger from chain data
-    RebuildLedger(rebuild_ledger::Args),
+    /// rebuilds ledger and chain from WAL
+    Rebuild(rebuild::Args),
```
scarmuega: lets call this command `RebuildStores`

src/serve/grpc/sync.rs
```diff
@@ -203,20 +216,24 @@ impl u5c::sync::sync_service_server::SyncService for SyncServiceImpl {
...
-            .map_err(|_err| Status::internal("can't read WAL"))?
-            .ok_or(Status::internal("WAL has no data"))?;
+        let (slot, body) = match self
+            .chain
+            .get_tip()
+            .map_err(|e| Status::internal(format!("Unable to read WAL: {:?}", e)))?
+        {
+            Some((slot, body)) => (slot, body),
+            None => return Err(Status::internal("chain has no data.")),
+        };
```
scarmuega: use `ok_or()` for better readability

src/serve/grpc/sync.rs
```diff
@@ -104,19 +108,28 @@ impl u5c::sync::sync_service_server::SyncService for SyncServiceImpl {
...
+                        let block = self.mapper.map_block_cbor(&body);
+
+                        Ok(u5c::sync::AnyChainBlock {
+                            native_bytes: body.to_vec().into(),
+                            chain: u5c::sync::any_chain_block::Chain::Cardano(block).into(),
+                        })
+                    }
+                    None => Err(Status::not_found(format!("Failed to find block: {:?}", br))),
+                }
+            })
```
scarmuega: can't we use here the `raw_to_anychain` method that we had before?

src/bin/dolos/data/export.rs
```diff
@@ -11,6 +11,10 @@ pub struct Args {
     /// the path to export to
     #[arg(short, long)]
     output: PathBuf,
+
+    // Whether to include ledger
+    #[arg(long, action)]
+    include_chain: bool,
```
scarmuega: The description seems wrong, but regardless, I think it would be a good idea to have a flag to include each store, chain and / or ledger in the export.

src/bin/dolos/data/prune_chain.rs
```diff
@@ -0,0 +1,48 @@
+use miette::{bail, Context, IntoDiagnostic};
+use tracing::info;
+
+#[derive(Debug, clap::Args)]
+pub struct Args {
+    /// the maximum number of slots to keep in the WAL
```
scarmuega: typo in description

src/bin/dolos/data/prune_wal.rs
```diff
@@ -15,11 +15,11 @@ pub struct Args {
 pub fn run(config: &crate::Config, args: &Args) -> miette::Result<()> {
     crate::common::setup_tracing(&config.logging)?;
 
-    let (mut wal, _) = crate::common::open_data_stores(config).context("opening data stores")?;
+    let mut wal = crate::common::open_wal(config).context("opening data stores")?;
 
     let max_slots = match args.max_slots {
         Some(x) => x,
-        None => match config.storage.max_wal_history {
+        None => match config.storage.max_chain_history {
```
scarmuega: shouldn't it be `config.storage.max_wal_history`?

src/ledger/mod.rs
```diff
@@ -126,8 +139,17 @@ pub fn compute_delta(
...
+    let era: u16 = block.era().into();
     let mut delta = LedgerDelta {
         new_position: Some(ChainPoint(block.slot(), block.hash())),
+        new_block: match block {
+            MultiEraBlock::Byron(x) => minicbor::to_vec((era, x)).unwrap(),
+            MultiEraBlock::Conway(x) => minicbor::to_vec((era, x)).unwrap(),
+            MultiEraBlock::Babbage(x) => minicbor::to_vec((era, x)).unwrap(),
+            MultiEraBlock::AlonzoCompatible(x, _) => minicbor::to_vec((era, x)).unwrap(),
+            MultiEraBlock::EpochBoundary(x) => minicbor::to_vec((0_u16, x)).unwrap(),
+            _ => Default::default(),
```
scarmuega: this encoding procedure should be moved to Pallas, but not a requirement for merging the PR.

In case this doesn't get migrated, I would prefer that the default case throws a todo!() instead of providing an empty vec.

src/ledger/mod.rs
```diff
@@ -170,8 +192,17 @@ pub fn compute_undo_delta(
...
+    let era: u16 = block.era().into();
     let mut delta = LedgerDelta {
         undone_position: Some(ChainPoint(block.slot(), block.hash())),
+        undone_block: match block {
+            MultiEraBlock::Byron(x) => minicbor::to_vec((era, x)).unwrap(),
+            MultiEraBlock::Conway(x) => minicbor::to_vec((era, x)).unwrap(),
+            MultiEraBlock::Babbage(x) => minicbor::to_vec((era, x)).unwrap(),
+            MultiEraBlock::AlonzoCompatible(x, _) => minicbor::to_vec((era, x)).unwrap(),
+            MultiEraBlock::EpochBoundary(x) => minicbor::to_vec((0_u16, x)).unwrap(),
+            _ => Default::default(),
```
scarmuega: same as prev comment.

src/serve/grpc/sync.rs
```diff
@@ -104,19 +110,20 @@ impl u5c::sync::sync_service_server::SyncService for SyncServiceImpl {
...
+            .iter()
+            .map(|br| {
+                match self
+                    .chain
+                    .get_block_by_slot(&br.index)
+                    .map_err(|_| Status::internal("Failed to query chain service."))?
+                {
+                    Some(body) => Ok(raw_to_anychain(&self.mapper, &body)),
+                    None => Err(Status::not_found(format!("Failed to find block: {:?}", br))),
+                }
```
scarmuega: prefer `ok_or()`

src/model.rs
```diff
@@ -43,8 +45,47 @@ pub struct SubmitConfig {
...
+        match repr {
+            Some(version) => match version.as_str() {
+                "v0" => Ok(StorageVersion::V0),
+                "v1" => Ok(StorageVersion::V1),
+                _ => Err(<D::Error as serde::de::Error>::custom("Invalid version")),
+            },
+            None => Ok(StorageVersion::V0),
+        }
+    }
+}
```
scarmuega, line 70: user `#[serde_as(as = "DisplayFromStr")]` instead. Example here:
https://github.com/txpipe/dolos/blob/ae1ade251a4ac446b81e8ad0d3074d343101696e/src/bin/dolos/main.rs#L101C5-L101C39

src/sync/mod.rs
```diff
@@ -64,11 +66,12 @@ pub fn pipeline(
         quit_on_tip,
     );
 
-    let mut roll = roll::Stage::new(wal.clone());
+    let mut roll = roll::Stage::new(wal.clone(), chain.clone());
```
scarmuega: roll stage should be agnostic of the chain, it should only care about the WAL.

src/sync/roll.rs
```diff
@@ -33,9 +35,10 @@ pub struct Stage {
 }
 
 impl Stage {
-    pub fn new(store: WalStore) -> Self {
+    pub fn new(store: WalStore, chain: ChainStore) -> Self {
```
scarmuega: no need for `chain` store in this stage

src/sync/roll.rs
```diff
@@ -110,7 +113,10 @@ impl gasket::framework::Worker<Stage> for Worker {
     async fn execute(&mut self, unit: &WorkUnit, stage: &mut Stage) -> Result<(), WorkerError> {
         match unit {
             WorkUnit::PullEvent(pull) => stage.process_pull_event(pull).await?,
-            WorkUnit::Housekeeping => stage.store.housekeeping().or_panic()?,
+            WorkUnit::Housekeeping => {
+                stage.store.housekeeping().or_panic()?;
+                stage.chain.housekeeping().or_panic()?;
```
scarmuega: housekeeping of chain should happen in the `apply` stage.

scarmuega (approved):

PR #453, opened 2025-02-27
fix: Rebuild ledger procedure

gonzalezzfelipe:
The rebuild ledger procedure was broken because it pulls from the WAL Apply blocks, but fails to apply the ledger delta of those that where undone


src/wal/reader.rs
```diff
@@ -162,3 +166,64 @@ pub trait WalReader: Clone {
...
+
+impl<T> WalBlockReader<'_, T>
+where
+    T: WalReader,
+{
+    pub fn try_new(wal: &T, start: Option<LogSeq>, genesis: &Genesis) -> Result<Self, WalError> {
+        let mut undone = BTreeSet::new();
+        let mut iter = wal.crawl_from(start)?;
+        let security_window = ((3.0 * genesis.byron.protocol_consts.k as f32)
+            / (genesis.shelley.active_slots_coeff.unwrap())) as u64;
```
scarmuega: move this value outside of the WalBlockReader and pass it as an arg

scarmuega (approved):

PR #457, opened 2025-03-06
fix: Dump history now uses WalBlockReader

### LINKED ISSUE #456: fix: implement WAL block reader (with look-ahead) for the dump history endpoint
The dump-history endpoint of the UTxO RPC server suffers from the same issue with undo records when reading a list of blocks from the WAL. We need to switch it to the same new WAL block reader with look-ahead capabilities.

gonzalezzfelipe:
Fixes #456

### DIFF
--- a/src/wal/reader.rs
+++ b/src/wal/reader.rs
@@ -1,5 +1,7 @@
 use std::collections::BTreeSet;
 
+use crate::ledger::pparams::Genesis;
+
 use super::*;
 
 pub trait ReadUtils<'a> {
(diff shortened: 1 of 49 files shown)


src/wal/reader.rs
```diff
@@ -213,6 +215,12 @@ where
             window: iter,
         })
     }
+
+    /// Extract lookahead window from Genesis.
+    pub fn lookahead_from_genesis(genesis: &Genesis) -> u64 {
+        ((3.0 * genesis.byron.protocol_consts.k as f32)
+            / (genesis.shelley.active_slots_coeff.unwrap())) as u64
+    }
```
scarmuega: doesn't belong here. The ledger should provide a function describing the "mutability window" from genesis data.

scarmuega: I think there's already a method in there.

scarmuega:
@gonzalezzfelipe we can close this one, right?
I believe it got deprecated by the new chain store.

PR #466, opened 2025-04-01
chore: Add server http (minibf) prompt

gonzalezzfelipe:


src/bin/dolos/init/mod.rs
```diff
@@ -185,6 +185,10 @@ pub struct Args {
     #[arg(long)]
     serve_grpc: Option<bool>,
 
+    /// Serve clients via HTTP
+    #[arg(long)]
+    serve_http: Option<bool>,
```
scarmuega: feature should be specific to blockfrost.

lets use `serve_minibf`

src/bin/dolos/init/mod.rs
```diff
@@ -380,6 +400,16 @@ impl ConfigEditor {
         Ok(self.apply_serve_grpc(Some(value)))
     }
 
+    fn prompt_serve_http(self) -> miette::Result<Self> {
+        let value = Confirm::new("Do you want to serve clients via HTTP (Blockfrost)?")
```
scarmuega: `Do you want to serve clients via a Blockfrost-like HTTP endpoint?`

scarmuega (approved):

PR #495, opened 2025-05-07
fix(trp): cost models index matches LanguageView

nicolasLuduena:
Protocol parameters' cost model index matches how LanguageView indexes them. All credit for the find goes to @alegadea


nicolasLuduena:
It depends on https://github.com/tx3-lang/tx3/pull/89 being merged as well.

scarmuega (approved):

PR #500, opened 2025-05-14
feat(serve): implement new ReadTx spec endpoint

lancevincentsalera:
Hello @scarmuega, @SebastienGllmt. This PR implements the introduced `ReadTx` endpoint in utxrpc/spec: https://github.com/utxorpc/spec/pull/143. I would very much appreciate your reviews and thoughts for this. Let me know if there's anything I need to change. Thank you!

**NOTE**: This includes changes from https://github.com/utxorpc/spec/pull/155 since `ReadTx` is directly affected with the inclusion of block information inside tx watchers


src/serve/grpc/query.rs
```diff
@@ -363,4 +369,55 @@ impl u5c::query::query_service_server::QueryService for QueryServiceImpl {
...
+        let message = request.into_inner();
+
+        info!("received new grpc query");
+
+        let tx_hash = message.hash;
+
+        let (block_body, tx) = self
+            .chain
+            .get_tx_with_block_data(&tx_hash)
+            .map_err(|e| Status::not_found(format!("transaction not found: {e}")))?
```
gonzalezzfelipe: I think it makes more sense for this to be a 500 instead of a 400, this would be thrown if there was a problem opening or querying the db
```suggestion
            .map_err(|e| Status::internal(format!("failed to query chain for tx: {e}")))?
```

lancevincentsalera: That makes sense. Thank you for this @gonzalezzfelipe !

lancevincentsalera: Resolved. Let me know if there's anything else!

gonzalezzfelipe (approved):
LGTM

src/chain/redb/v1.rs
```diff
@@ -311,11 +311,22 @@ impl ChainStore {
...
+        if let Some((_, tx)) = tx_with_block {
+            return Ok(Some(tx));
+        }
+        Ok(None)
+    }
+
+    pub fn get_tx_with_block_data(
+        &self,
+        tx_hash: &[u8],
+    ) -> Result<Option<(BlockBody, Vec<u8>)>, Error> {
```
SebastienGllmt: I think `BlockBody` used everywhere in this PR is the wrong name. A block is very loosely split into two things:
1. block header
2. block body

`MultiEraBlock` consists of the full block (body, header, anything else). That's also what I think is logical to return from utxorpc as well. However, in this PR, you refer to the fully block as "BlockBody" which is misleading because it makes it sound like you're returning only the "body" field of the block (whereas you're actually returning the full block)

Although I think this terminology overloading stems from beyond this PR, since I also this this ambiguity in types like `RawBlock`

lancevincentsalera: Makes sense. Thank you for pointing this out @SebastienGllmt ! Already applied the [changes](https://github.com/txpipe/dolos/pull/500/commits/8f7b453fdfe3b646428df739fd9aa30b1063a7ec).

SebastienGllmt: maybe just to loop in @scarmuega since I'm not sure if there was a separate rationale for calling things "body"

PR #502, opened 2025-05-15
feat(serve): implement ReadChainConfig spec endpoint

lancevincentsalera:
## NOTE:
- `genesis` field of `ReadChainConfigResponse` is currently returned as an empty byte array.
- EraSummary.config is currently an empty string ("").

We’ll need to follow up on how best to populate these in Dolos

CC: @SebastienGllmt


src/serve/grpc/query.rs
```diff
@@ -363,4 +398,59 @@ impl u5c::query::query_service_server::QueryService for QueryServiceImpl {
...
+            .collect();
+
+        let network_id = match self.genesis.shelley.network_id.as_ref() {
+            Some(network) => network.clone(),
+            None => "".to_string(),
+        };
+        let network_magic = self.genesis.shelley.network_magic.unwrap();
+        let caip2 = format!("cip34:{}-{}", network_id, network_magic);
+
+        let genesis = "".to_string();
```
SebastienGllmt, line 437: use the genesis block hash. For example, on mainnet, I believe the genesis block hash is `5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb`

src/serve/grpc/query.rs
```diff
@@ -215,6 +218,38 @@ fn into_u5c_utxo(
...
+            time: start.timestamp.timestamp() as u64,
+            slot: start.slot,
+            epoch: start.epoch,
+        }),
+        end: Some(u5c::cardano::EraBoundary {
+            time: end.timestamp.timestamp() as u64,
+            slot: end.slot,
+            epoch: end.epoch,
+        }),
+        config: "".to_string(),
```
SebastienGllmt, line 238: This should be the JSON for the era. In Dolos, you can get this from `byron_path`

PR #504, opened 2025-05-19
fix(serve): handle new spec block-structure changes

lancevincentsalera:
This PR addresses spec block structure changes from https://github.com/utxorpc/spec/pull/155 and https://github.com/utxorpc/spec/pull/154. 

CC: @SebastienGllmt, @scarmuega


src/serve/grpc/watch.rs
```diff
@@ -229,7 +240,7 @@ impl u5c::watch::watch_service_server::WatchService for WatchServiceImpl {
         let intersect = inner_req
             .intersect
             .iter()
-            .map(|x| ChainPoint::Specific(x.index, x.hash.to_vec().as_slice().into()))
+            .map(|x| ChainPoint::Specific(x.slot, x.hash.to_vec().as_slice().into()))
```
SebastienGllmt, line 243: This PR properly adds the changes in the output, but isn't handling the new behavior in the input. The goal of #154 is that you should be able to create requests on more than just the slot number (since you might only know the block height, or only know the block hash)

lancevincentsalera: Hello @SebastienGllmt , thanks for this. I made changes to the fetchBlock endpoint, can you check if the implementation has more to improve or looks good already? I tested this on my side and it works fine now. Many thanks!

SebastienGllmt: yes, LGTM!

PR #523, opened 2025-06-06
added otel with once_cell

Yiakman:
This is an updated version of this [PR by SebastienGllmt](https://github.com/txpipe/dolos/pull/400)
Main objective is to consume the logs by Paima Engine


src/bin/dolos/main.rs
```diff
@@ -116,6 +131,12 @@ pub struct LoggingConfig {
 
     #[serde(default)]
     include_minibf: bool,
+
+    #[serde(default)]
```
Yiakman, line 135: This adds two new fields for the logging, a boolean for setting up telemetry and telemetry settings for mainly specifying the endpoint to send logs

src/bin/dolos/common.rs
```diff
@@ -130,26 +139,86 @@ pub fn setup_tracing(config: &LoggingConfig) -> miette::Result<()> {
...
+                .with(console_subscriber::spawn())
+                .with(filter)
+                .init();
+        }
     }
 
     Ok(())
 }
 
+pub fn shutdown_tracing() {
```
Yiakman, line 216: graceful shutdown for telemetry. Its invoked when the daemon is shutting down

src/bin/dolos/common.rs
```diff
@@ -130,26 +139,86 @@ pub fn setup_tracing(config: &LoggingConfig) -> miette::Result<()> {
...
+                    .with_attribute(KeyValue::new(
+                        "service.version",
+                        env!("CARGO_PKG_VERSION"),
+                    ))
+                    .build(),
+            )
+            .build();
+
+        // keep provider alive for the whole program
+        LOGGER_PROVIDER.set(provider.clone()).ok();
```
Yiakman, line 171: If I didn't do this the provider went out of scoped and it was dropped. This keeps it running while the program is running and allows to invoke the shutdown afterwards

PR #545, opened 2025-06-17
fix: Make UntilTip sync option reachable

gonzalezzfelipe:
As shown [here](https://github.com/txpipe/pallas/blob/main/pallas-network/src/miniprotocols/chainsync/client.rs#L255-L267), `NetworkResponse::Await` is never returned by the `recv_while_must_reply`, so the `stage.quota.on_tip()` is never called.


src/sync/pull.rs
```diff
@@ -167,6 +167,7 @@ impl gasket::framework::Worker<Stage> for Worker {
             debug!("should request next batch of blocks");
             Ok(WorkSchedule::Unit(WorkUnit::Pull))
         } else {
+            stage.quota.on_tip();
```
scarmuega: Having side-effects on the schedule call is weird. This should go in the execute method.

PR #591, opened 2025-07-10
feat: add OTLP exporter

gonzalezzfelipe:


gonzalezzfelipe:
extends https://github.com/txpipe/dolos/pull/523

PR #627, opened 2025-07-30
fix: Fix pagination and error handling for bf conformance

gonzalezzfelipe:


crates/minibf/src/routes/addresses.rs
```diff
@@ -14,24 +14,41 @@ use dolos_cardano::pparams::ChainSummary;
...
+    // If the address is not seen on the chain, send 404.
+    if domain
+        .archive()
+        .iter_blocks_with_address(&address.to_vec())
+        .iter()
+        .next()
+        .is_none()
+    {
+        return Err(Error::Code(StatusCode::NOT_FOUND));
+    }
```
scarmuega, line 51: I think we could skip the 404 check if `!refs.is_empty()`

PR #648, opened 2025-08-15
feat(grpc): implement Read Era Summary and Read Genesis

hrmtm23:
- Implement Read Genesis and Read Era Summary
- Modified block structure: added timestamp, and modified name index to slot

## Summary by CodeRabbit

* **New Features**
  * CAIP-2 network identifiers added to API responses.
  * Ledger tip and block references now include computed timestamps.

* **Bug Fixes**
  * More accurate block heights and timestamps across tip info, genesis/era summaries, and watch/transaction streams using era-aware protocol parameter calculations.
  * Added robust fallback handling for missing or undecodable block data to improve reliability.

### DIFF
--- a/src/serve/grpc/query.rs
+++ b/src/serve/grpc/query.rs
@@ -40,57 +40,55 @@ where
         Self { domain, mapper }
     }
 
-    fn point_to_u5c(&self, point: &ChainPoint) -> u5c::query::ChainPoint {
+    fn point_to_u5c(&self, point: &ChainPoint) -> Result<u5c::query::ChainPoint, Status> {
         match point {
-            ChainPoint::Origin => u5c::query::ChainPoint {
+            ChainPoint::Origin => Ok(u5c::query::ChainPoint {
                 slot: 0,
                 hash: vec![].into(),
                 height: 0,
                 timestamp: 0,
-            },
+            }),
             ChainPoint::Slot(slot) | ChainPoint::Specific(slot, _) => {
                 // Calculate height by looking up block from slot
-                let height = self.domain
+                let body = self.domain
                     .archive()
                     .get_block_by_slot(slot)
-                    .map(|block| {
-                        block.map(|body| {
-                            // Parse the block to get the height
-                            use pallas::ledger::traverse::MultiEraBlock;
-                            if let Ok(parsed_block) = MultiEraBlock::decode(&body) {
-                                parsed_block.number()
-                            } else {
-                                *slot  // Fallback to slot
-                            }
-                        }).unwrap_or(*slot)
-                    })
-                    .unwrap_or(*slot);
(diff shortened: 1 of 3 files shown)


src/serve/grpc/query.rs
```diff
@@ -34,6 +39,61 @@ where
...
+                            // Parse the block to get the height
+                            use pallas::ledger::traverse::MultiEraBlock;
+                            if let Ok(parsed_block) = MultiEraBlock::decode(&body) {
+                                parsed_block.number()
+                            } else {
+                                *slot  // Fallback to slot
+                            }
+                        }).unwrap_or(*slot)
+                    })
+                    .unwrap_or(*slot);
```
gonzalezzfelipe: I don't think it is a good idea to return the block slot instead of the block height if the block isn't parseable or it can't be found. Instead we should return `Status::internal`

src/serve/grpc/query.rs
```diff
@@ -34,6 +39,61 @@ where
...
+                let updates = self.domain.state()
+                    .get_pparams(*slot)
+                    .ok()
+                    .and_then(|updates| {
+                        updates.into_iter()
+                            .map(TryInto::try_into)
+                            .collect::<Result<Vec<_>, _>>()
+                            .ok()
+                    })
+                    .unwrap_or_default();
```
gonzalezzfelipe: Same here, instead of making an empty vec if the pparams are unresolved we should return a `Status::internal`

src/serve/grpc/query.rs
```diff
@@ -418,23 +478,99 @@ where
...
+
+        // Get genesis hash
+        let genesis_hash = if let Ok(Some(point)) = self.domain.state().cursor() {
+            match point {
+                ChainPoint::Origin | ChainPoint::Slot(_) => vec![],
+                ChainPoint::Specific(_, hash) => hash.to_vec(),
+            }
+        } else {
+            vec![]
+        };
```
gonzalezzfelipe: Is this right? Shouldn't this be a static hash?

gonzalezzfelipe (changes requested):

src/serve/grpc/query.rs
hrmtm23: Fixed - Replaced all slot fallbacks with proper error returns.

src/serve/grpc/query.rs
hrmtm23: Fixed - Now uses static `genesis.shelley_hash`

gonzalezzfelipe (commented):
@hrmtm23 its looking good, sory for the delayed response. Some of the APIs changed a bit but the functionality you are using is still available. Could you rebase and handle the conflict resolution?

PR #657, opened 2025-08-25
feat: implement readtx grpc

paulobressan:
Closes #500

### DIFF
--- a/crates/testing/src/toy_domain.rs
+++ b/crates/testing/src/toy_domain.rs
@@ -154,10 +154,10 @@ impl pallas::interop::utxorpc::LedgerContext for ToyDomain {
         &self,
         _refs: &[pallas::interop::utxorpc::TxoRef],
     ) -> Option<pallas::interop::utxorpc::UtxoMap> {
-        todo!()
+        None
     }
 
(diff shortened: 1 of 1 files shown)


crates/testing/src/toy_domain.rs
```diff
@@ -148,3 +148,16 @@ impl dolos_core::Domain for ToyDomain {
...
+        &self,
+        _refs: &[pallas::interop::utxorpc::TxoRef],
+    ) -> Option<pallas::interop::utxorpc::UtxoMap> {
+        todo!()
+    }
+
+    fn get_slot_timestamp(&self, _slot: u64) -> Option<u64> {
+        todo!()
+    }
+}
```
scarmuega: we should be able to implement these methods for the toy domain. Or at least, return some dummy data instead of panicking.

scarmuega (approved):

crates/testing/src/toy_domain.rs
paulobressan: The latest implementation had `todo!()`. I changed it to return `None` for now so it doesn't break the test. We can explore a mock ledger when needed.

PR #664, opened 2025-09-05
fix(compiler): include tip_slot in Compiler configuration

sofia-bobbiesi:


crates/trp/src/compiler.rs
```diff
@@ -50,18 +50,24 @@ fn build_pparams<D: Domain>(domain: &D) -> Result<tx3_cardano::PParams, Error> {
...
+    let hash = cursor.hash().map(|h| h.to_vec()).unwrap_or_default();
+
+    let tip = tx3_cardano::ChainPoint { slot, hash };
+
     let compiler = tx3_cardano::Compiler::new(
         pparams,
         tx3_cardano::Config {
             extra_fees: config.extra_fees,
         },
+        tip,
```
gonzalezzfelipe, line 70: Some changes seem to be missing on the Compiler side in order to accept this new argument, is this PR dependent on some other release?

PR #733, opened 2025-09-29
Support for writing validity range using posix time

sofia-bobbiesi:
## Summary by CodeRabbit

* **Bug Fixes**
  * Added a specific error for unresolved tip resolution and a dedicated error code.
  * Compiler initialization now includes chain point (slot/hash/timestamp) for accurate tip handling.

* **Tests**
  * Adjusted test protocol text formatting.


crates/trp/src/compiler.rs
```diff
@@ -59,18 +59,33 @@ fn build_pparams<D: Domain>(domain: &D) -> Result<tx3_cardano::PParams, Error> {
...
+        hash,
+        timestamp,
+    };
+
     let compiler = tx3_cardano::Compiler::new(
         pparams,
         tx3_cardano::Config {
             extra_fees: config.extra_fees,
         },
+        tip,
```
gonzalezzfelipe, line 76: Is this awaiting a pending release on tx3_cardano to work?

PR #756, opened 2025-10-12
feat(utxorpc): add block to WatchTx

nelsonksh:
## Summary by CodeRabbit

* **New Features**
  * Transaction watch stream can now optionally include the full originating block when requested, providing richer context for consumers.
  * Cardano block data and native serialized bytes are embedded alongside transactions for improved traceability and analysis.

* **Chore**
  * Backward-compatible: existing clients remain unaffected unless they request the extra block payload.


gonzalezzfelipe (approved):

nelsonksh:
Closing as superseded. The core of this PR, returning the block alongside each matched tx in `WatchTx`, landed in #833 ("include correlating block in Watch matches") and was carried into both the `v1alpha` and `v1beta` watch services by #987. It has shipped in releases since v1.0.0.

What's still not upstream is the field-mask opt-out that lets callers who don't need the block skip it (the concern raised above). Sync got that in #1012 (`masking.rs`), but Watch always embeds the block. If that's wanted for Watch, a small fresh PR following the #1012 pattern would be cleaner than rebasing this one.

Thanks for the reviews!

PR #822, opened 2025-12-22
fix(cardano): Handle pool re-registrations

gonzalezzfelipe:
## Summary by CodeRabbit

* **New Features**
  * Automatic tracking and queuing of pool-delegator retirement events so retiring delegations are processed at epoch boundaries.
  * Account state now preserves a retired-pool value for fallback display and logic when no active delegation exists.
  * Account view/build logic falls back to the preserved retired pool when determining a delegator's pool.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/crates/cardano/src/ewrap/drops.rs
+++ b/crates/cardano/src/ewrap/drops.rs
@@ -158,13 +158,13 @@ impl super::BoundaryVisitor for BoundaryVisitor {
     ) -> Result<(), ChainError> {
         let current_epoch = ctx.ending_state.number;
 
+        // Notice that instead of dropping delegators when a pool is retired, we're moving the data to a different field to be able to still track the relationsihp between the pool and the delegators.
+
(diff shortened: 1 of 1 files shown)


crates/cardano/src/model.rs
```diff
@@ -533,6 +533,9 @@ pub struct AccountState {
 
     #[n(6)]
     pub credential: StakeCredential,
+
+    #[n(7)]
+    pub retired_pool: Option<PoolHash>,
```
scarmuega, line 539: let's put a cbor default attribute here so that previous snapshots still work regardless of the existence of the new value.

gonzalezzfelipe: resolved at [d9b758c](https://github.com/txpipe/dolos/pull/822/commits/d9b758cbab91096878fb12a104a5970ec01f47c8)

PR #833, opened 2026-01-18
Fix `outputs_match_asset` to not require asset_name

SupernaviX:
The `outputs_match_asset` util wouldn't allow users to match policy ids unless they were also matching asset names. Update it to support this.

Also, fix the `Phase1ValidationRejected` error message so that it tells users what's wrong with their TX.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages for transaction validation rejections to include additional context.
  * Fixed block information handling to ensure complete block data is properly preserved during transaction processing.
  * Enhanced asset matching logic for more precise filtering by policy and asset name.


SupernaviX:
> @SupernaviX thanks for the contribution, functionality wise I think it makes perfect sense. Can you handle the conflicts for us to go forward with a merge?

I merged the main branch in. Tests are failing but they were before, and I didn't see any merge conflicts

PR #872, opened 2026-02-06
feat(minibf): add optional base_path configuration

### LINKED ISSUE #1298: feat(minibf): add optional base_path configuration
Synthetic Kanban issue for:
- #872 by @adrian1-dot

adrian1-dot:
Resolves #1298.

### Problem

The [Blockfrost OpenAPI specification](https://docs.blockfrost.io/) defines all API endpoints under the `/api/v0` base path (see `servers` section in the official spec). However, Dolos minibf currently serves endpoints directly at the root level (e.g., `/network`, `/blocks/latest`), which breaks compatibility with tools and libraries that expect canonical Blockfrost endpoint paths.

This incompatibility causes integration issues with:
- **[blockfrost-haskell](https://github.com/blockfrost/blockfrost-haskell)**: Client library with hardcoded `/api/v0` base path
- Other Blockfrost-compatible tooling may also expect standard endpoint structure

### Solution

Add an optional `base_path` configuration field to `MinibfConfig` that allows nesting all routes under a custom prefix.

**Configuration example:**
```toml
[serve.minibf]
listen_address = "[::]:3000"
base_path = "/api/v0"  # Optional - omit for root-level routes
```

When configured:
- All routes become accessible at `{base_path}/{endpoint}` (e.g., `/api/v0/network`)
- Original root-level routes are no longer accessible

### Changes

1. **crates/core/src/config.rs**: Added `base_path: Option<String>` field to `MinibfConfig`
2. **crates/minibf/src/lib.rs**: Modified `build_router()` to conditionally nest all routes when `base_path` is set (~7 lines)
3. **crates/minibf/README.md**: Added documentation for the new configuration option
4. **docs/content/apis/minibf.mdx**: Updated API documentation
5. **src/bin/dolos/init.rs**: Added `base_path` to default configuration initialization

### Implementation Details

The implementation uses Axum's `Router::nest()` to wrap the entire route tree under the custom base path when configured:

```rust
if let Some(base_path) = &cfg.base_path {
    Router::new().nest(base_path, app).layer(NormalizePathLayer::trim_trailing_slash())
} else {
    app.layer(NormalizePathLayer::trim_trailing_slash())
}
```

This approach minimizes code changes while providing complete Blockfrost spec compliance.

### Backwards Compatibility

- [x] **Fully backwards compatible** - when `base_path` is not configured, all routes remain at root level exactly as before.

### Testing

- [x] All 110 existing tests pass (`cargo test --workspace --all-features`)
- [x] No clippy warnings (`cargo clippy --workspace --all-features`)
- [x] Successfully tested with local [atlas](https://github.com/geniusyield/atlas) server using `/api/v0` base path
- [x] Verified endpoints accessible at `/api/v0/*` with base_path configured
- [x] Verified endpoints at root level when base_path is omitted

### Impact

This change enables Dolos minibf to serve as a drop-in replacement for Blockfrost API in tools requiring canonical endpoint paths, significantly improving ecosystem compatibility while maintaining flexibility for users who prefer root-level routes.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable base_path option to prefix all minibf HTTP endpoints (e.g., serve under /api/v0).

* **Documentation**
  * Updated docs and README with examples showing base_path usage and backward-compatibility when omitted.

* **Bug Fixes / Improvements**
  * Base path values are validated and produce clear configuration errors if invalid.
  * Improved runtime diagnostics when router construction fails.

[Review Change Stack](https://app.coderabbit.ai/change-stack/txpipe/dolos/pull/872?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


crates/minibf/src/lib.rs
```diff
@@ -449,18 +422,37 @@ where
...
         });
-    app.layer(NormalizePathLayer::trim_trailing_slash())
+    
+        // Optionally nest all routes under base_path
+        if let Some(base_path) = &cfg.base_path {
+            // Validate before using
+            if base_path.is_empty() || base_path == "/" || !base_path.starts_with('/') || base_path.contains('*') {
+                panic!(
+                    "base_path must start with '/', must not be just '/', and must not contain wildcards; got: \"{}\"",
+                    base_path
```
scarmuega: we should avoid a panic and instead bubble up an error, ideally a "config" error variant.

adrian1-dot: I've addressed the panic by adding a `ConfigError(String)` variant to `ServeError` in `dolos-core` and changing `build_router` to return `Result<Router, ServeError>`. Invalid `base_path` values now return `Err(ServeError::ConfigError(...))` instead of panicking, propagating through `Driver::run` consistently with `BindError` and `ShutdownError`.

I've also rebased onto the current `main` to resolve the CI conflicts.

scarmuega:
Thanks for this — Blockfrost `/api/v0` spec compliance is worth having. The branch has gone stale (the minibf crate was restructured, so there are extensive merge conflicts). Could you rebase onto current `main`? The core change is small enough that it should rebase cleanly once the router code is re-pointed. One open question for us: whether `base_path` should stay opt-in (as here) or whether we eventually default to `/api/v0` for closer Blockfrost parity — for now opt-in is the right call.

adrian1-dot:
Rebased onto current `main` and squashed to a single commit. The `minibf` restructuring resolved cleanly — diff is now ~60 lines focused on `MinibfConfig.base_path` + `ServeError::ConfigError`.

The inline `ConfigError` change from the earlier review is preserved: invalid `base_path` values (empty, `/` alone, missing leading `/`, or containing `*`) return `Err(ServeError::ConfigError(...))` instead of panicking.

On your open question — agreed that opt-in is the right call for now. Defaulting to `/api/v0` would silently break every existing root-level integration; happy to revisit in a follow-up once there's a deprecation path.

adrian1-dot:
Addressed both coderabbit findings in `8f61088`:

- **Trailing-slash canonicalization**: `base_path` is now trimmed of trailing `/` before validation/`Router::nest`, so `"/api/v0/"` and `"/api/v0"` behave identically.
- **Integration tests**: added `base_path_tests` covering (1) prefixed routes resolve + root 404s when `base_path` is set, (2) trailing-slash input is normalized, (3) invalid inputs (`""`, `"/"`, missing leading `/`, contains `*`) return `ServeError::ConfigError`.

All 214 minibf tests pass.

crates/minibf/src/lib.rs
```diff
@@ -507,7 +508,24 @@ where
...
+                "base_path must start with '/', must not be just '/', and must not contain wildcards; got: \"{}\"",
+                base_path
+            )));
+        }
+        Ok(Router::new()
+            .nest(base_path, app)
+            .layer(NormalizePathLayer::trim_trailing_slash()))
+    } else {
+        Ok(app.layer(NormalizePathLayer::trim_trailing_slash()))
+    }
```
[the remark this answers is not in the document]
adrian1-dot: resolved both https://github.com/txpipe/dolos/pull/872#issuecomment-4529945824

crates/minibf/src/lib.rs
```diff
@@ -658,7 +659,25 @@ where
         } else {
             CorsLayer::new()
         });
-    app.layer(NormalizePathLayer::trim_trailing_slash())
+
+    if let Some(base_path) = &base_path {
+        let base_path = base_path.trim_end_matches('/');
+        if base_path.is_empty()
+            || base_path == "/"
```
michalrus: Dead condition, `== "/"` will never match because `trim_end_matches('/')` already removed that slash.

crates/minibf/src/lib.rs
```diff
@@ -658,7 +659,25 @@ where
...
+                "base_path must start with '/', must not be just '/', and must not contain wildcards; got: \"{}\"",
+                base_path
+            )));
+        }
+        Ok(Router::new()
+            .nest(base_path, app)
+            .layer(NormalizePathLayer::trim_trailing_slash()))
+    } else {
+        Ok(app.layer(NormalizePathLayer::trim_trailing_slash()))
+    }
```
michalrus: `NormalizePathLayer` is attached with `Router::layer`, so it runs *after* route matching and never rewrites the path before routing.

Verified against this router: `/api/v0/network/` and the bare `/api/v0/` return 404 instead of being normalized.

crates/minibf/src/lib.rs
```diff
@@ -687,3 +706,49 @@ where
...
+        let (status, _) = app.get_bytes("/network").await;
+        assert_eq!(
+            status,
+            StatusCode::NOT_FOUND,
+            "root route should 404 when base_path is set"
+        );
+    }
+
+    #[tokio::test]
+    async fn trailing_slash_in_base_path_is_normalized() {
```
michalrus: Slightly misleading test name. This only proves that a trailing slash in the *config value* (`/api/v0/`) is trimmed before nesting. It never issues a trailing-slash *request*, which currently fails, see the previous comment about `NormalizePathLayer`.

src/bin/dolos/minibf.rs
```diff
@@ -34,7 +34,9 @@ pub async fn run(config: &RootConfig, args: &Args) -> miette::Result<()> {
         .into_diagnostic()
         .context("invalid minibf path")?;
 
-    let app = dolos_minibf::build_router(minibf.clone(), domain);
+    let app = dolos_minibf::build_router(minibf.clone(), domain)
+        .into_diagnostic()
+        .context("building minibf router")?;
```
michalrus, line 54: `build_router` now nests every route under `cfg.base_path`, so this local query subcommand silently requires the prefix in `args.path` (e.g. `dolos minibf /api/v0/network`).

A bare `dolos minibf /network` will 404 whenever `base_path` is configured.

Let's prepend `base_path` to `path` here so the CLI stays agnostic to the server prefix?

michalrus (changes requested):

crates/minibf/src/lib.rs
michalrus: Fixed in 9c52e02f85293be13709647d4bb607c1e1a71561.

crates/minibf/src/lib.rs
michalrus: Fixed in 9c52e02f85293be13709647d4bb607c1e1a71561.

crates/minibf/src/lib.rs
michalrus: Fixed in 9c52e02f85293be13709647d4bb607c1e1a71561.

src/bin/dolos/minibf.rs
michalrus: Fixed in 9c52e02f85293be13709647d4bb607c1e1a71561.

crates/core/src/lib.rs
```diff
@@ -337,6 +337,9 @@ impl WalError {
 
 #[derive(Debug, Error)]
 pub enum ServeError {
+    #[error("invalid configuration: {0}")]
+    ConfigError(String),
+
```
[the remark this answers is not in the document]
michalrus: @scarmuega, what do you think?

crates/minibf/src/lib.rs
```diff
@@ -341,7 +342,7 @@ impl<D: Domain> Facade<D> {
 
 pub struct Driver;
 
-pub fn build_router<D>(cfg: MinibfConfig, domain: D) -> Router
+pub fn build_router<D>(cfg: MinibfConfig, domain: D) -> Result<Router, ServeError>
 where
```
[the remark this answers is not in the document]
michalrus: @scarmuega, what do you think?

michalrus (approved):

crates/core/src/lib.rs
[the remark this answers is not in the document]
scarmuega: Ideally, config errors should be handled at config-parse time, not throw from inside a runtime process like "serve". I prefer not to expand this enum.

crates/minibf/src/lib.rs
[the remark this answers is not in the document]
scarmuega: building the router should be a non-fallible operation. Pre-condition checks belong somewhere else.

crates/core/src/lib.rs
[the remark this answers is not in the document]
michalrus: Fixed in 06eacdf6ace4aba922dc1096489d7e337736aa4a.

crates/minibf/src/lib.rs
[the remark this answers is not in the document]
michalrus: Fixed in 06eacdf6ace4aba922dc1096489d7e337736aa4a.

src/bin/dolos/main.rs
```diff
@@ -88,7 +88,14 @@ fn main() -> Result<()> {
 
     let config = crate::common::load_config(&args.config)
         .into_diagnostic()
-        .context("parsing configuration");
+        .context("parsing configuration")
+        .and_then(|config| {
+            config
+                .validate()
+                .map_err(|e| miette::miette!("{e}"))
+                .context("validating configuration")?;
```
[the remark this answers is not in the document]
michalrus: Fixed in f74155b8fc407ca3a715b86d0ef8f495609674f0.

src/bin/dolos/minibf.rs
```diff
@@ -29,6 +29,20 @@ pub async fn run(config: &RootConfig, args: &Args) -> miette::Result<()> {
...
+    let path = match minibf.base_path() {
+        Some(base_path) => {
+            if path == base_path || path.starts_with(&format!("{base_path}/")) {
+                path
+            } else {
+                format!("{base_path}{path}")
+            }
+        }
+        None => path,
+    };
```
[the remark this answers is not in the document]
michalrus: Fixed in f74155b8fc407ca3a715b86d0ef8f495609674f0.

crates/core/src/config.rs
```diff
@@ -819,6 +825,30 @@ impl MinibfConfig {
...
+        if let Some(base_path) = self.base_path() {
+            if base_path.is_empty() || !base_path.starts_with('/') || base_path.contains('*') {
+                return Err(format!(
+                    "base_path \"{base_path}\" is not valid. Use a base_path that starts with '/' and has no '*' wildcard."
+                ));
+            }
+        }
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
michalrus: Fixed in f74155b8fc407ca3a715b86d0ef8f495609674f0.

crates/minibf/src/lib.rs
```diff
@@ -690,7 +692,23 @@ where
...
+
+    // NormalizePath must receive the request before the router matches a route.
+    // Then NormalizePath can remove a trailing slash from the path. A layer
+    // added with `Router::layer` runs after the route match. Thus, a request
+    // with a trailing slash does not match a route. The code sets the normalized
+    // router as the fallback service of an empty router. As a result, every request
+    // goes to this fallback service, and the public return type remains
+    // `Router`.
+    let normalized = NormalizePathLayer::trim_trailing_slash().layer(router);
+    Router::new().fallback_service(normalized)
```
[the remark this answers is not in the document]
michalrus: I will keep this implementation. Axum 0.8.4 documents that `Router::layer` runs after routing. Thus, it cannot normalize the URI before route matching. The fallback-service wrapper applies `NormalizePathLayer` before routing.

crates/core/src/config.rs
```diff
@@ -819,6 +825,32 @@ impl MinibfConfig {
...
+    /// does not stop a running `serve` process.
+    pub fn validate(&self) -> Result<(), String> {
+        if let Some(configured_base_path) = self.base_path.as_deref() {
+            let base_path = configured_base_path.trim_end_matches('/');
+            if base_path.is_empty() || !base_path.starts_with('/') || base_path.contains('*') {
+                return Err(format!(
+                    "base_path \"{configured_base_path}\" is not valid. Use a base_path that starts with '/' and has no '*' wildcard."
+                ));
+            }
+        }
```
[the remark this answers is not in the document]
michalrus: Fixed in bb98adc169427eed815ec588e2eff0e678f5090b.

michalrus (approved):

crates/core/src/config.rs
```diff
@@ -819,6 +825,36 @@ impl MinibfConfig {
...
+    /// Rejects a `base_path` that the router cannot nest under. This check runs
+    /// at config-parse time. As a result, a malformed value stops startup and
+    /// does not stop a running `serve` process.
+    pub fn validate(&self) -> Result<(), String> {
+        if let Some(configured_base_path) = self.base_path.as_deref() {
+            let base_path = configured_base_path.trim_end_matches('/');
+            if base_path.is_empty()
+                || !base_path.starts_with('/')
+                || base_path.contains(['*', '?', '#'])
+                || base_path.chars().any(char::is_whitespace)
```
michalrus: The doc comment above promises "the nest operation cannot fail", but this blocklist does not cover axum/matchit route syntax. Verified: `/api/{`, `/api/}`, `/api/{}` and `/api/:v0` all pass this check and then panic inside `Router::nest` at startup ("Parameters must be registered with a valid name" / "Path segments must not start with `:`"). Worse, `/api/{v}` passes AND builds: it becomes a capture group, so `/api/ANYTHING/network` resolves, and every handler using `Path<T>` returns 500 ("Wrong number of path arguments").

michalrus (changes requested):

michalrus (changes requested):
Also, this table should contain the new `base_path` field:

https://github.com/txpipe/dolos/blob/cd971fc9dd8bf8298d220ea5fb5d06be3be8d732/docs/content/configuration/schema.mdx?plain=1#L300-L310

crates/core/src/config.rs
michalrus: A panicking driver takes the daemon down (`monitor_drivers` propagates it), so a typo in `dolos.toml` is a crash, not a config error.

michalrus: Fixed in ec0281133e4a978b213636d50ddcd2b61aadc769.

michalrus (approved):
↑ Both fixed in ec0281133e4a978b213636d50ddcd2b61aadc769.

PR #976, opened 2026-04-21
chore(minibf): Implement /pools/pool_id endpoint

gonzalezzfelipe:
## Release Notes

* **New Features**
  * Added CIP-151 pool registration metadata parsing and validation for transaction data
  * Introduced pool certificates indexing dimension for historical archive queries
  * Implemented new `/pools/{id}` endpoint returning detailed pool information including live/active stake metrics, registration/retirement history, and CIP151 Calidus key support
  * Added pool retirement certificate detection and parsing
  * Extended Bech32 encoding to support Calidus key serialization

### DIFF
--- a/crates/minibf/src/routes/pools.rs
+++ b/crates/minibf/src/routes/pools.rs
@@ -6,37 +6,337 @@ use axum::{
     Json,
 };
 use blockfrost_openapi::models::{
-    pool_delegators_inner::PoolDelegatorsInner, pool_history_inner::PoolHistoryInner,
+    pool::Pool, pool_delegators_inner::PoolDelegatorsInner, pool_history_inner::PoolHistoryInner,
     pool_list_extended_inner::PoolListExtendedInner, PoolListExtendedInnerMetadata,
 };
 use dolos_cardano::{
+    indexes::{AsyncCardanoQueryExt, SlotOrder},
     model::{AccountState, PoolState},
-    PoolDelegation, StakeLog,
+    pallas_extras, PoolDelegation, PoolHash, StakeLog,
 };
 use dolos_core::{BlockSlot, Domain, EntityKey};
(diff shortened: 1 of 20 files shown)


crates/minibf/src/routes/pools.rs
```diff
@@ -6,37 +6,337 @@ use axum::{
...
+        active_size: safe_ratio(metrics.active_stake, metrics.total_active_stake),
+        declared_pledge: params.pledge.to_string(),
+        live_pledge: metrics.live_pledge.to_string(),
+        margin_cost: rational_to_f64::<6>(&params.margin),
+        fixed_cost: params.cost.to_string(),
+        reward_account,
+        owners,
+        registration,
+        retirement,
+        calidus_key: None,
```
gonzalezzfelipe: `calidus_key` lives on tx metadata, on label 867. We have 2 options:

* search for it using our tx metadata index. We scan all blocks with the 867 label on a transaction and try and find the calidus key for the particular block. If there is no calidus key, we go through every single block with that label. Probably not that expensive, but still ugly.
* add it to PoolState in the same way that we are handling the asset metadata, with something like `#[cbor(default)] calidus_key_slot: Option<BlockSlot>`. We implement a visitor that extracts this value, validates its format, and if everything is right we put it in the field. Then on runtime we just go get that block.

@scarmuega what do you think is better? IMO second one is clearly superior but it implies modifying the state. BUT first one is easier, and we are already paying a high price on this endpoint because we have to scan all accounts to calculate live metrics.

scarmuega: we'll go for the scan approach and we'll revisit these performance optimizations as a cross-cut concern at a later stage.

scarmuega (approved):

scarmuega (approved):

PR #1000, opened 2026-05-22
fix(grpc): populate block_ref in AnyUtxoData via block_by_tx_hash

islamaliev:
## What

`into_u5c_utxo` (in both `src/serve/grpc/v1alpha/query.rs` and `src/serve/grpc/v1beta/query.rs`) now populates `AnyUtxoData.block_ref` instead of returning `None`. Every UTxO returned by `read_utxos` and `search_utxos` will now carry the block (slot/hash/height/timestamp) that produced it.

## Why

The `block_ref` field was added to `AnyUtxoData` as part of the u5c 0.18.1 upgrade in #813 (commit `2b6de0bc`), but it landed as a placeholder — `block_ref: None` — and the reverse lookup was never wired up. Consumers querying UTxOs through `read_utxos` / `search_utxos` therefore lose all block-level metadata for those UTxOs.

A downstream library ([`pogun-chainfollower`](https://github.com/input-output-hk/pogun-chainfollower/pull/212)) shipped a `filter_map` tourniquet to silently drop these items so the rest of the pipeline doesn't fall over. Populating the field at the source makes that workaround unnecessary.

## How

Mirror the pattern already used by `read_tx` (a few functions down in the same file): call `AsyncQueryFacade::block_by_tx_hash(txo.0.to_vec())`, decode the block with `MultiEraBlock::decode`, and build a `ChainPoint { slot, hash, height, timestamp }`. The reverse lookup is backed by the existing `slot_by_tx_hash` index — no new indexing is needed; it is already implemented across both storage backends (`dolos-redb3`, `dolos-fjall`).

Lookup failures (tx not in archive, decode error, query error) `warn!` with the tx hash and fall back to `block_ref: None`. This mirrors the existing posture in `into_u5c_utxo` for datum-decode errors — graceful degradation rather than failing the whole call.

Applied to both `v1alpha` and `v1beta` (near-identical files).

## Performance tradeoff (please weigh in)

`into_u5c_utxo` is invoked once per UTxO. With this patch, a `search_utxos` page of N UTxOs costs N reverse-lookups + N full `MultiEraBlock::decode` calls. When the UTxOs cluster on M < N source txs (typical — multi-output txs), N − M of those decodes are redundant.

This PR takes the **Tier 1** path: per-UTxO lookup, with a `TODO` comment near the new code marking the spot for future batching. Correctness over performance, acceptable cost for typical query sizes.

**Tier 2** (deferred): refactor `read_utxos` / `search_utxos` to gather unique tx hashes from the page first, batch-fetch a `HashMap<TxHash, ChainPoint>`, and thread it into `into_u5c_utxo` via a signature change. Bigger diff, better cost profile.

Happy to land Tier 2 inline before merge if reviewers prefer it, or to track it as a follow-up issue — deferring to maintainer judgment.

## Tests

No existing test harness exercises `read_utxos` / `search_utxos` against fixture blocks — the in-file test module only covers genesis/era-summary RPCs against the in-memory `ToyDomain`, which doesn't have meaningful archive integration for `block_by_tx_hash` to return `Some`. I didn't want to invent a heavy fixture/seeded-chain scaffolding without first asking whether maintainers want one. **Please advise**:

- comfortable merging on review-only and tracking a test in a follow-up?
- or would you like a test (and is there a preferred harness pattern — e2e under `tests/` against a seeded snapshot, or a unit test with a richer mock domain)?

## Quality gates (locally, against `a5dc18f`)

- `cargo fmt --check` — the two touched files are clean. (Pre-existing fmt diffs in unrelated files exist on `main` because `rustfmt.toml` enables `wrap_comments = true`, which only applies on nightly. Untouched.)
- `cargo build --workspace --all-targets --all-features` — clean.
- `cargo test --workspace --all-targets` — 540 passed, 0 failed (matches CI shape — `ci.yml` does not pass `--all-features`).
  - Note: `cargo test --workspace --all-features` shows 14 pre-existing failures in `dolos-cardano` proptests (`model::accounts::prop_tests::*`, `model::epochs::prop_tests::*`, `model::pools::prop_tests::pool_registration_roundtrip`, etc.). I verified these fail on bare `main@a5dc18f` without my changes, so they're unrelated to this PR.

## Scope

- Both `v1alpha` and `v1beta` updated.
- No proto changes (`block_ref: Option<ChainPoint>` already exists on `AnyUtxoData`).
- No new storage indexes.
- No unrelated refactors.

## Summary by CodeRabbit

* **New Features**
  * UTXO query responses now include block reference metadata (slot, block hash, block height, transaction index) for returned UTXOs.
  * The system prefetches and deduplicates block lookups to reduce redundant work when resolving UTXO origins.
* **Bug Fixes**
  * Missing or malformed block data is handled gracefully with warnings; queries continue when some block refs are unavailable.

[Review Change Stack](https://app.coderabbit.ai/change-stack/txpipe/dolos/pull/1000?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/core/src/async_query.rs
+++ b/crates/core/src/async_query.rs
@@ -5,9 +5,22 @@ use tokio::sync::Semaphore;
 use pallas::ledger::traverse::MultiEraBlock;
 
 use crate::{
-    archive::ArchiveStore, indexes::IndexStore, ArchiveError, BlockBody, BlockSlot, ChainError,
-    ChainPoint, Domain, DomainError, EraCbor, IndexError, TagDimension, TxOrder,
+    archive::ArchiveStore, indexes::IndexStore, ArchiveError, BlockBody, BlockHash, BlockHeight,
+    BlockSlot, ChainError, ChainPoint, Domain, DomainError, EraCbor, IndexError, TagDimension,
+    TxOrder,
 };
+
+/// Lightweight block metadata for a transaction, extracted via a single decode.
+///
+/// Returned by `block_meta_by_tx_hash`. Callers that need the full block body
+/// should use `block_by_tx_hash` instead.
+#[derive(Debug, Clone)]
+pub struct BlockRefMeta {
+    pub slot: BlockSlot,
+    pub hash: BlockHash,
+    pub height: BlockHeight,
+    pub tx_index: TxOrder,
+}
 
 #[derive(Debug, Clone)]
 pub struct AsyncQueryOptions {
@@ -132,6 +145,42 @@ where
         }
 
         Ok(None)
+    }
+
+    /// Look up the block containing a given transaction hash and return only
+    /// chain-point metadata, decoding the block once inside the blocking task.
+    ///
+    /// Prefer this over `block_by_tx_hash` when only the chain point is needed —
+    /// it avoids a second `MultiEraBlock::decode` in the caller.
+    pub async fn block_meta_by_tx_hash(
+        &self,
+        tx_hash: Vec<u8>,
+    ) -> Result<Option<BlockRefMeta>, DomainError> {
+        self.run_blocking(move |domain| {
+            let Some(slot) = domain.indexes().slot_by_tx_hash(&tx_hash)? else {
+                return Ok(None);
+            };
+            let Some(raw) = domain.archive().get_block_by_slot(&slot)? else {
+                return Ok(None);
+            };
+            let block = MultiEraBlock::decode(raw.as_slice())
+                .map_err(|e| DomainError::ChainError(ChainError::DecodingError(e)))?;
+            let Some((tx_index, _)) = block
+                .txs()
+                .iter()
+                .enumerate()
+                .find(|(_, tx)| tx.hash().as_slice() == tx_hash.as_slice())
+            else {
+                return Ok(None);
+            };
+            Ok(Some(BlockRefMeta {
+                slot: block.slot(),
+                hash: block.hash(),
+                height: block.number(),
+                tx_index,
+            }))
+        })
+        .await
     }
 
     pub async fn tx_cbor(&self, tx_hash: Vec<u8>) -> Result<Option<EraCbor>, DomainError> {
--- a/src/serve/grpc/block_refs.rs
+++ b/src/serve/grpc/block_refs.rs
@@ -1,0 +1,97 @@
+//! Shared helpers for resolving `block_ref` metadata for UTxOs returned by
+//! `read_utxos` / `search_utxos` across the v1alpha and v1beta query services.
+//!
+//! Loads era summary once per request and deduplicates `block_meta_by_tx_hash`
+//! lookups by source transaction — UTxOs sharing a producing tx pay a single
+//! decode rather than one per item.
+
+use std::collections::HashMap;
+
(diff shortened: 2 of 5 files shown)


islamaliev:
Pushed a follow-up commit ([b3b1495](https://github.com/txpipe/dolos/pull/1000/commits/b3b1495)) addressing the bulk of the review feedback on the original Tier-1 patch:

- **A-01 (double decode)** — added `AsyncQueryFacade::block_meta_by_tx_hash` returning a chain-point-only `BlockRefMeta` decoded once inside the blocking task. The gRPC caller no longer re-decodes the same block.
- **A-02 (era-summary reload)** — `load_era_summary` is now called once per request, not once per UTxO. Threaded through `BlockRefData::timestamp`.
- **A-03 (dedupe per page)** — Tier 2 batching is in. The new `serve::grpc::block_refs::fetch_block_refs` collects unique source tx hashes from the UTxO page, performs one lookup per unique tx, and returns a `HashMap<TxHash, BlockRefData>` that callers index per-item. A `search_utxos` page whose 100 UTxOs cluster on 20 source txs now does 20 lookups, not 100.
- **A-04 (silent error swallowing)** — storage errors from `block_meta_by_tx_hash` now propagate as `Status::internal`. `Ok(None)` (tx not in archive) stays silent as before — that's still a legitimate state for not-yet-indexed tips and pruned/custom UTxOs.
- **A-05 (log amplification)** — per-UTxO `warn!` is gone; block-decode failures (indicating archive corruption) drop to `debug!` instead.
- **A-06 (v1alpha/v1beta duplication)** — the whole flow now lives in `src/serve/grpc/block_refs.rs`. Each version-specific `query.rs` only ships a 7-line `to_chain_point(&BlockRefData) -> u5c::query::ChainPoint` helper to convert into its own proto type.
- **C-01 (nested match)** — fixed as a side effect of the helper extraction; the gRPC `query.rs` files no longer carry the lookup logic at all.
- **C-02** (allocation pattern) — withdrawn after verifying `TxHash = Hash<32>` is `Copy`; the original code was idiomatic.

The two `into_u5c_utxo` functions now just receive `Option<ChainPoint>` as a parameter — no per-UTxO lookup, no per-UTxO storage hits, no per-UTxO warns.

Local quality gates against the new HEAD:
- `cargo build --workspace --all-targets --all-features` — clean.
- `cargo test --workspace --all-targets` — 540 passed, 0 failed (matches CI shape).
- Touched files are `cargo fmt --check`-clean.

Tests still deferred per the original PR question. The reviewer noted that `minibf::TestDomainBuilder` (seeding `ToyDomain` via `import_blocks`) plus `FaultyToyDomain` for error branches would cover the happy path + archive-miss + storage-error cases in ~30–40 lines. Happy to add them in this PR if you prefer.

gonzalezzfelipe (approved):
@scarmuega LGTM. As a follow up, we should take this approach and standardize it, using it also on https://github.com/txpipe/dolos/blob/main/crates/minibf/src/mapping.rs#L381-L386 and https://github.com/txpipe/dolos/blob/main/crates/minikupo/src/routes/matches.rs#L232-L236

scarmuega (approved):

PR #1017, opened 2026-06-17
fix(cardano): fail loud on lagging pool snapshots and unfinished epoch boundaries

scarmuega:
> **Scope note:** this is **hardening / fail-loud**, not recovery. It does **not** implement true shard resume and does **not** repair an already-lagging pool. Root cause (a persisted pool snapshot lagging the epoch) is still unconfirmed — most likely non-idempotent mid-boundary crash resume, flagged by the existing `TODO: implement true shard resume` (restored here) and tracked in #1018.

## Context

Follow-up to #1016, which turned an obscure `unwrap` panic at RUPD into a clear error when a `PoolState` snapshot lags the current epoch. That only *surfaced* the condition. This PR adds two **fail-loud guards** so the same corruption is caught earlier and a half-finished boundary can't silently double-apply.

### What we established
- The healthy boundary path **cannot** produce a lag: ESTART `commit_finalize` advances `EpochState.number` in the **same atomic commit** as every `PoolTransition` + cursor.
- A persisted lag therefore comes from an abnormal crash/resume. The sharding code (#978) already flags this with **`TODO: implement true shard resume`** and the note that **`AccountTransition` is not natively idempotent**. That is the suspected root cause; it is **not fixed here** (see #1018).

### Piece A — guard the silent-corruption hole
`MintedBlocksInc::apply` accumulates blocks into the pool's positional `live` slot, which only holds this epoch's blocks when the snapshot is aligned. `apply` now **asserts** alignment to the block's epoch before accumulating — a descriptive panic at the origin, consistent with the infallible delta-apply layer's existing invariant `expect`s. The block epoch rides on a transient `#[serde(skip)]` field; WAL deltas are only ever undone, so the WAL format is unchanged.

### Piece B — guard ESTART finalize
`commit_finalize` now asserts every shard committed and the epoch hasn't advanced before rotating pools / advancing the epoch, returning `BrokenInvariant::EpochBoundaryIncomplete` otherwise. A **defensive assertion on the finalize step only** — it does **not** make the per-shard `AccountTransition` replay idempotent (the restored shard-resume TODO stands).

### Error codes + troubleshooting
Both errors are now **codified** with concise messages and documented in a new `docs/content/operations/troubleshooting.mdx`:
- **`CARDANO-001`** — pool snapshot lagging (RUPD path + the `MintedBlocksInc::apply` assert)
- **`CARDANO-002`** — epoch boundary incomplete (ESTART finalize guard)

The verbose explanation/operator guidance moved from the message strings into that page.

## Explicitly NOT in this PR
- **True shard resume** — idempotent mid-boundary resume (the actual fix; #1018).
- **Repair of already-corrupted state** — nodes that already persisted a lag keep failing loud with `CARDANO-001` and need a re-bootstrap.

## Testing
- `cargo clippy --workspace --all-targets` — clean.
- `cargo test -p dolos-cardano --lib` — 122 pass, incl. `minting_into_lagging_pool_panics` (`#[should_panic(expected = "CARDANO-001")]`) and an aligned `minted_blocks_inc_roundtrip`.
- Piece B's positive path is exercised by the boundary-crossing suites (`epoch_pots`, `tests/cardano`) in CI; fixtures are unavailable locally (dangling symlinks, pre-existing).

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened epoch-boundary resume to ensure finalize proceeds only when boundary work is fully complete, preventing unwanted era advancement, streaming, archive logging, and cursor updates.
  * Added epoch-aware validation for minted-block deltas to detect pool snapshot lag and reject mismatched-epoch updates with a clear error.
  * Improved epoch-boundary and crash-recovery diagnostics with more precise error reporting, including committed/total details.
* **Documentation**
  * Expanded troubleshooting docs with structured guidance for CARDANO-001 through CARDANO-005.
* **Tests**
  * Updated and extended minted-block delta coverage, including a regression test for epoch-mismatch panics.

### DIFF
--- a/crates/cardano/src/estart/commit.rs
+++ b/crates/cardano/src/estart/commit.rs
@@ -184,6 +184,23 @@ impl super::WorkContext {
     ) -> Result<(), ChainError> {
         debug!("committing estart finalize changes");
 
+        // Finalize advances the epoch and rotates every pool in one commit, so
+        // it must run exactly once and only after all shards committed. Require
+        // `committed == total` and that the epoch has not advanced, turning a
+        // would-be double-rotation into a loud error. (Guards finalize only,
+        // not per-shard replay — see the "true shard resume" TODO.)
+        let ended = self.ended_state();
+        let progress = ended.estart_progress.as_ref();
+        let all_shards_committed = progress.is_some_and(|p| p.committed == p.total);
+        if !all_shards_committed {
+            return Err(dolos_core::BrokenInvariant::EpochBoundaryIncomplete {
+                epoch: ended.number,
+                committed: progress.map(|p| p.committed),
+                total: progress.map(|p| p.total),
+            }
+            .into());
+        }
+
         // Collect era transition data first (only 1-2 entities, not a memory concern)
         let era_transition = self.collect_era_transition(state)?;
 
--- a/crates/cardano/src/estart/work_unit.rs
+++ b/crates/cardano/src/estart/work_unit.rs
@@ -15,7 +15,8 @@
 //!
 //! `AccountTransition` is not natively idempotent (re-applying double-rolls
 //! the EpochValue snapshot), so true mid-shard resume requires additional
-//! work — same TODO posture as `EwrapWorkUnit`.
+//! work — same TODO posture as `EwrapWorkUnit`. `finalize` is separately
+//! guarded to run exactly once (see `commit_finalize`).
 
 use std::sync::Arc;
 
--- a/crates/cardano/src/lib.rs
+++ b/crates/cardano/src/lib.rs
@@ -357,105 +357,51 @@ impl dolos_core::ChainLogic for CardanoLogic {
             None => WorkBuffer::Empty,
         };
 
-        // Crash-recovery check: if the previous process crashed mid-boundary,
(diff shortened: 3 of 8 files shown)


scarmuega:
Thanks @coderabbitai — evaluated all three:

1. **Cache `ended_state()` (`commit.rs`)** — applied. Pure field borrow, cleaner.
2. **"Assert" vs "Check" wording** — already addressed in a later commit; the comment now reads "Require `committed == total` …".
3. **Missing mutual-exclusion fail-loud (`lib.rs`) — declined, and the comment that prompted it was fixed.** `ewrap_progress` and `estart_progress` are **not** mutually exclusive. Per the `EpochWrapUpV3` contract, after the close phase `ewrap_progress` is deliberately left at `Some(total, total)` as the "EWRAP done, ESTART pending finalize" marker and is only cleared by `EpochTransition` at ESTART finalize. So during the entire ESTART (open) phase **both** fields are `Some` by design — returning `BrokenInvariant` when both are set would crash every healthy node that restarts mid-ESTART. The real defect was my own comment claiming "exactly one is set"; I've corrected it to describe the V3 lingering-marker behavior, and the two independent checks (which can both legitimately fire on one restart) stay as-is.

Issue #1018, opened 2026-06-17
Epoch boundary resume is not idempotent — root cause of persisted pool-snapshot lag


scarmuega:
## Summary

A node restart (Mithril-bootstrapped preview, "WAL in sync with state") panicked at RUPD for epoch 1223 because at least one persisted `PoolState` had `snapshot.epoch() != EpochState.number` — its snapshot lagged the current epoch. #1016 turned that obscure `unwrap` panic into a clear `ChainError::PoolSnapshotLagging`; #1017 added two fail-loud guards. **Neither fixes the root cause.** This issue captures what we know and what the real fix is.

## What we established (by code reading)

1. **The healthy boundary path cannot persist a lag.** `EpochState.number` advances only in ESTART `commit_finalize` (`crates/cardano/src/estart/commit.rs`), via `EpochTransition`, in the **same atomic `writer.commit()`** as every `PoolTransition` and the cursor. Both backends (`redb3`, `fjall`) commit a `Writer` atomically. So pools and the epoch advance together-or-not-at-all. The import path (`crates/core/src/import.rs`) runs the same lifecycle.

2. **Therefore a persisted lag originates in an abnormal crash/resume.** The sharding work (#978) already flags this directly in `crates/cardano/src/lib.rs`:
   > correctness depends on shard idempotency … **`AccountTransition` is not natively idempotent** … Operators should **monitor the subsequent boundary for inconsistency**. **TODO: implement true shard resume.**

   The boundary transitions (`AccountTransition`, `PoolTransition`, `EpochTransition`, `NonceTransition`, DRep/Proposal transitions) call `default_transition(next_epoch)` **unconditionally** — re-applying one double-rotates the `EpochValue` snapshot window and increments the epoch again. The only thing preventing re-application today is the **shard-level** skip (`start_shard` reading `estart_progress.committed`) plus #1017's finalize guard. The authors flagged the shard-level protection as insufficient, and they were closer to the code than our from-reading inference.

3. **Atomicity ≠ idempotency.** Atomic per-shard commit (verified) prevents torn writes *within* a shard. It does **not** guarantee that re-running a boundary from a partially-committed state reconstructs correct state — that is the missing property.

## Open question: lead vs lag

The simplest non-idempotency story (re-running a transition) produces an epoch **lead** (entity ahead). The reported symptom is a **lag** (pool behind `EpochState`), i.e. the pool was transitioned *fewer* times than the epoch. Candidate lag mechanisms to confirm:
[...]

PR #1048, opened 2026-07-06
fix(bootstrap): recover state from WAL when WAL is ahead after crash

scarmuega:
## Problem

A crash between `commit_wal` and `commit_state` (which write to separate redb databases, so they are not atomic) leaves the **WAL tip ahead of the state cursor** — WAL at block N, state at N-1.

On restart, this was silently unrecoverable and corrupted the ledger:
- `catch_up_stores` only brought archive/index up to the **state** cursor (N-1), never to the WAL tip.
- The upstream puller builds intersection candidates from the **WAL tip** (`intersect_candidates`, `src/sync/pull.rs`), so the peer resumed streaming at **N+1**.
- `receive_block` does no continuity check against the state cursor, so N+1 applied cleanly on top of N-1 — block **N's effects (produced/consumed UTxOs) were never applied to state and were silently lost**, and the corruption compounds forward.

## Root cause

Recovery treated **state as the leader** and only reconciled followers *down to* it, while the sync intersection trusts the WAL. Nothing brought state *up to* the WAL after a crash in that window.

## Fix

- Retarget `catch_up_stores` so **every** store (state, archive, index) reconciles forward to the **WAL tip** under one uniform rule.
- Add `catch_up_state`: forward-replays WAL entries into the state store — the exact mirror of the existing `rollback` loop (load entity → `delta.apply` → `save_entity_typed`), plus `apply_utxoset(compute_catchup(..).utxo_delta)` and cursor advance, committed per entry.
- **Why replay is lossless:** only `Roll`/`Genesis` work units write the WAL (the default `commit_wal` is a no-op; RUPD/EWRAP/ESTART mutate state directly). So a WAL that is ahead of state can only ever be ahead by `Roll` blocks, and each `Roll` WAL entry fully captures its own state mutation. Recovering a crash *mid*-epoch-boundary is out of scope here — that's the separate #1018 gap (noted in code).
- Set the state cursor to `Origin` at the end of genesis `execute` (set last, so a mid-genesis crash re-runs genesis cleanly). This lets restart distinguish "genesis done, no blocks yet" from "genesis never ran", so it doesn't re-run genesis — whose `commit_wal` resets the WAL — over blocks awaiting recovery.
- Skip synthetic empty-block WAL entries (from `reset_to`) in all catch-up paths.

### Behavior note

Because genesis now leaves the state cursor at `Origin` instead of `None`, `dolos bootstrap`'s `has_existing_data` reports **true after a genesis-only run**, so re-bootstrapping requires `--force`/`--skip-if-data`. This is intentional and safer than silently importing over genesis-seeded state.

## Testing

- New `test_catchup_recovers_state_from_wal` (WAL-only feed → `bootstrap()` → state/archive/index all at WAL tip; the replayed block's UTxO is queryable from state) and `test_catchup_converges_all_stores_to_wal_tip` (each store at a different point behind the tip).
- Passing: `dolos-core` (122), `dolos-cardano` (16), `tests/bootstrap.rs` (5, incl. existing catch-up/rollback/origin regressions), plus `sync`/`housekeeping`/`snapshot` integration tests; workspace check + clippy clean.
- `tests/boundary_resume.rs` (#1019) is not on `main`, so that regression check should be re-run once the branches meet.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved crash recovery by tightening WAL boundary handling, correctly distinguishing “genesis ran but no blocks yet” from “genesis never ran,” and preventing unintended re-initialization.
  * Reworked state, archive, and index catch-up to replay precisely up to the WAL tip while ignoring synthetic reset placeholders, then added verification so boot fails on residual gaps.
  * Refined rollback to undo and persist entity changes more consistently before truncating history.
* **Tests**
  * Expanded bootstrap and rollback coverage to simulate partial commits and store divergence, asserting convergence to the WAL tip and byte-identical rollback state.

### DIFF
--- a/crates/core/src/bootstrap.rs
+++ b/crates/core/src/bootstrap.rs
@@ -232,6 +232,25 @@ fn catch_up_state<D: Domain>(domain: &D, target: &ChainPoint) -> Result<(), Doma
         info!(count, "state caught up from WAL");
     }
 
+    // Post-condition: the replay must actually reach the target. A WAL that
+    // claims a tip the state can't reach (e.g. entries wiped by `reset_to`)
+    // is unrecoverable here — fail loudly instead of leaving a silent gap.
+    // Compared by slot: a Slot-only cursor at the target's slot (post-ESTART)
+    // is already at the target even though the points differ.
+    let cursor = domain.state().read_cursor()?;
+
+    let reached = cursor
+        .as_ref()
+        .is_some_and(|cursor| cursor.slot() >= target.slot());
+
+    if !reached {
+        error!(?cursor, %target, "state catch-up could not reach the WAL target");
+        return Err(DomainError::InconsistentState {
+            wal: Some(target.clone()),
+            state: cursor,
+        });
+    }
+
     Ok(())
 }
 
@@ -274,6 +293,19 @@ fn catch_up_archive<D: Domain>(domain: &D, target: &ChainPoint) -> Result<(), Do
     if count > 0 {
         writer.commit()?;
         info!(count, "archive caught up from WAL");
+    }
+
+    // Archive can legitimately remain behind when the WAL no longer holds the
+    // missing blocks (e.g. after `reset_to`). Flag it instead of failing —
+    // archive completeness is not consensus-critical, matching the lenient
+    // handling in `check_archive_in_sync_with_state`.
+    let archive_tip = domain.archive().get_tip()?.map(|(slot, _)| slot);
+
+    if archive_tip < Some(target_slot) {
+        error!(
+            ?archive_tip,
+            target_slot, "archive still behind WAL target after catch-up"
+        );
     }
 
     Ok(())
@@ -320,6 +352,15 @@ fn catch_up_indexes<D: Domain>(domain: &D, target: &ChainPoint) -> Result<(), Do
     if count > 0 {
         writer.commit()?;
         info!(count, "indexes caught up from WAL");
+    }
+
+    // Same lenient handling as archive: flag a residual lag instead of
+    // failing the boot.
+    let index_cursor = domain.indexes().cursor()?;
+    let index_slot = index_cursor.as_ref().map(|p| p.slot());
+
+    if index_slot < Some(target.slot()) {
+        error!(?index_cursor, %target, "indexes still behind WAL target after catch-up");
     }
 
     Ok(())
--- a/crates/core/src/bootstrap.rs
+++ b/crates/core/src/bootstrap.rs
@@ -96,8 +96,13 @@ fn check_wal_in_sync_with_state<D: Domain>(domain: &D) -> Result<(), DomainError
             });
         }
         (Some(ref wal_tip), None) if wal_tip == &ChainPoint::Origin => {
-            // WAL at Origin with no state cursor — fresh node after genesis.
-            info!(%wal_tip, "WAL at origin, no state cursor (post-genesis)");
+            // WAL reset to Origin but no state cursor — a crash mid-genesis
+            // (genesis `commit_wal` resets the WAL to Origin before
+            // `commit_state` writes the cursor). A completed genesis leaves
+            // the cursor at Origin, so it takes the `(Some, Some)` arm above;
+            // this arm only fires when genesis was interrupted. Genesis
+            // re-runs cleanly on the next block.
+            info!(%wal_tip, "WAL at origin, no state cursor (interrupted genesis)");
         }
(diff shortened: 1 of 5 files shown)


tests/bootstrap.rs
```diff
@@ -4,36 +4,46 @@
...
+/// commit_archive/commit_indexes/finalize are always skipped.
+#[derive(Clone, Copy)]
+enum CrashAfter {
+    /// Run commit_wal only — models a crash between `commit_wal` and
+    /// `commit_state`.
+    Wal,
+    /// Run commit_wal + commit_state — models a crash between the state
+    /// commit and the archive/index commits.
+    State,
+}
```
scarmuega, line 38: what about the other potential crash locations? do we have test coverage for those?

crates/core/src/bootstrap.rs
```diff
@@ -138,34 +140,125 @@ fn check_archive_in_sync_with_state<D: Domain>(domain: &D) -> Result<(), DomainE
...
+
+        count += 1;
+    }
+
+    if count > 0 {
+        info!(count, "state caught up from WAL");
+    }
 
     Ok(())
 }
```
scarmuega, line 294: this code should be very similar to what the roll-forward does during normal operation. Is there an opportunity to abstract this in a function somewhere and make it more DRY?

tests/bootstrap.rs
scarmuega: Good catch — the matrix had a hole. There are three inter-store commit boundaries (`wal→state→archive→indexes`) and only two were modeled. Addressed in af02a336:

- `CrashAfter` now has `Wal` / `State` / `Archive` variants, one per boundary, and `test_catchup_recovers_indexes_when_archive_ahead` covers the previously untested archive→indexes window.
- Two crash classes are intentionally **not** covered here and now documented on the enum: crashes *inside* a phase (mid-shard) and crashes during epoch-boundary work units (RUPD/EWRAP/ESTART) — those don't write the WAL at all, so they can't be recovered by WAL replay. That's the #1018 boundary-resume territory.

crates/core/src/bootstrap.rs
scarmuega: Extracted in a3ee096f — with one correction to the framing: the true twin of this code isn't roll-forward but **rollback** (`sync.rs`), which runs the same load→mutate loop in the undo direction. Normal roll-forward (`Roll::commit_state`) persists entities already mutated in memory during `compute` — it never re-applies deltas — so only the save step overlaps there.

New shared kernels in `core/state.rs`: `apply_delta_chunk` / `undo_delta_chunk` (load missing entities into a caller-owned map, replay deltas forward/reverse) and `save_entities` (persist the map). `catch_up_state` uses the apply direction; `rollback` the undo direction. Commit cadence stays with each caller (catch-up commits per entry, rollback once) since they need different atomicity.

The refactor also surfaced and fixed a latent bug (was tracked in a local note): rollback undid entities in memory but **never saved them back** — post-rollback entity state still reflected the undone blocks. Rollback now threads one entity map across all undone entries (so consecutive blocks touching the same entity unwind from the in-memory value, not the stale uncommitted store) and persists it before commit. The rollback lifecycle test now asserts the accounts namespace is byte-identical to its snapshot at the rollback target.

crates/core/src/bootstrap.rs
```diff
@@ -138,34 +140,125 @@ fn check_archive_in_sync_with_state<D: Domain>(domain: &D) -> Result<(), DomainE
...
+    for (point, mut log) in logs {
+        // Skip entries at or before the current state cursor
+        if Some(point.slot()) <= state_slot {
+            continue;
+        }
+
+        // Skip synthetic entries (from reset_to) — they carry no effects
+        if log.block.is_empty() {
+            continue;
+        }
```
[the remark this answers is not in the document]
scarmuega: Valid core point — adopted with adjustments in 89494803:

- **State** (consensus-critical): hard post-condition. After replay, if the cursor doesn't reach the target slot, return `InconsistentState` instead of booting with a silent gap (e.g. WAL entries wiped by `reset_to`).
- **Archive/indexes**: log an explicit `error!` on residual lag but keep booting — matching the existing lenient philosophy of `check_archive_in_sync_with_state` (archive completeness isn't consensus-critical, and a hard fail would brick previously bootable nodes).
- One correction to the suggested check: comparison must be by **slot**, not point equality — a `Slot`-only cursor (post-ESTART) at the target's slot is legitimately at the target even though the points differ; point equality would false-positive there.

PR #1049, opened 2026-07-06
harden(sync): add apply-side block-sequence continuity guard

scarmuega:
## Summary

The sync pipeline had exactly **one** sequence-integrity check: header prev-hash/slot continuity in the pull stage's  (in-memory, reset on every pull-worker restart). Everything downstream was blind trust — fetched bodies were never validated against the persisted state cursor, so a duplicate, stale, or fork-sibling block was applied silently and only surfaced later as a confusing `input not found` error (the 4,436,119 incident).

This PR implements **Task 1** from `todo/hardening-block-sequence-integrity.md`: an apply-side continuity guard that closes the door on silent state poisoning.

### Guard: `RollWorkUnit::load()`
- Reads the state cursor **fresh** from the state store (satisfies post-rollback correctness — the guard checks against the real persisted position, not a stale in-memory value)
- Sorts the batch by slot, then calls `WorkBatch::check_continuity(cursor)`
- Fires **before** UTxO resolution, entity loading, or any state mutation — no poisoning possible
- Covers both sync and import paths (both flow through `execute_work_unit → load()`)

### Shared continuity rules (`consensus::check_extension`)
Per review feedback, the rules are implemented **once**, in `crates/cardano/src/consensus.rs`, and enforced at both layers:
- **Pull stage** (`ChainFragment::roll_forward`): per-header vs. the in-memory tip; failure → peer reconnect
- **Apply stage** (`WorkBatch::check_continuity`): folds `check_extension` over the sorted batch starting from the persisted cursor (or `Origin`), so intra-batch links are validated too; failure → abort before any commit

Rules: parent hash must match when both sides are known (skipped for genesis, Byron EBBs, and `Slot`-only cursors from Estart); slot must not regress (`>=` to allow same-slot Byron EBBs).

### Error handling
Also per review feedback, `ConsensusError` (`BrokenContinuity`, `SlotNotIncreasing`) lives in **dolos-cardano** — the rules are Cardano-specific and core stays chain-agnostic. It bubbles up through a generic boxed `ChainError::Consensus(Box<dyn Error>)` variant with a `ChainError::consensus()` constructor, following the existing `WalError::internal` pattern.

### Performance
- **Zero double-decoding** — `OwnedMultiEraBlock::decode` runs once in `receive_block`; the guard accesses `previous_hash()` on the already-decoded instance (field access, not a re-parse)
- **Cursor read:** one `state.read_cursor()` per batch (single key-value lookup), dwarfed by the existing `load_utxos` + `decode_utxos` that follow in the same `load()` call
- **Per-block cost:** one field access + one 32-byte hash compare + one slot compare. No allocations.

### Test infrastructure fixes
- `make_conway_block_with_prev(slot, prev_hash, block_number)` — new test helper for constructing chained blocks
- `build_synthetic_blocks` — fixed `sample_block` to chain `prev_hash` properly (was hardcoded `[9u8; 32]` for all blocks, which the new guard correctly caught)

## Tests

`tests/e2e/sync.rs` (target `sync`) — integration tests exercising `ToyDomain` as a black box via `roll_forward()` / `rollback()` (the same `SyncExt` entry point the apply stage uses):

| Test | Scenario |
|------|----------|
| `apply_rejects_non_contiguous_block` | Wrong parent hash → `BrokenContinuity`, cursor unchanged |
| `apply_rejects_slot_regression` | Slot goes backward → `SlotNotIncreasing`, cursor unchanged |
| `fork_simulation_applies_canonical_after_rollback` | The 4,436,119 scenario: apply orphan, rollback, apply canonical sibling |
| `continuity_check_skipped_for_genesis` | First block with `prev_hash=None` from Origin |
| `continuity_check_allows_same_slot_ebb` | `>=` slot semantics for Byron EBBs |
| `batch_internal_continuity_validated` | Chained blocks A→B→C apply cleanly; broken intra-batch link rejected |

Plus direct unit tests for `check_extension` in `consensus.rs`, covering the `Slot`-only-cursor cases (post-Estart) that can't be reached through `ChainFragment`.

## Verification

```bash
cargo clippy --workspace --all-targets                  # clean
cargo test -p dolos-core -p dolos-cardano               # 145 passed
cargo test --test sync                                  # 6 passed
cargo test --test bootstrap                             # 6 passed
cargo test --test housekeeping                          # 10 passed
```

## Related

- `todo/hardening-block-sequence-integrity.md` — Task 1 (this PR); Tasks 2-5 are follow-ups
- `todo/bug-relay-drops-rollback.md` — root cause of the 4,436,119 incident (relay-side fix is separate)
- `todo/bug-stale-workbuffer-after-rollback.md` — adjacent rollback-handling defect

## Summary by CodeRabbit

* **New Features**
  * Added stronger sync/roll-forward integrity checks to verify incoming blocks properly connect to the current state cursor and that time/slot progression is consistent before processing continues.

* **Bug Fixes**
  * Sync now rejects broken parent-hash chains and slot regressions with clearer, specific consensus errors, and preserves the state cursor when validation fails.
  * Improved correctness around rollback/fork handling, genesis transition, and accepted same-slot (Byron-style) sequences.

* **Tests**
  * Extended end-to-end and synthetic coverage to validate both passing and failing continuity scenarios.


Cargo.toml
```diff
@@ -143,6 +143,10 @@ path = "tests/mempool.rs"
 name = "bootstrap"
 path = "tests/bootstrap.rs"
 
+[[test]]
+name = "sync_integrity"
```
scarmuega: call it just `sync`

PR #1055, opened 2026-07-09
fix(minibf): return 404 for valid but non-existent asset in `/assets/{asset}/addresses`

### LINKED ISSUE #1119: minibf: fix `/assets/{asset}/addresses` to return 404 for valid but non-existent asset
The `/assets/{asset}/addresses` endpoint scanned the UTxO set for the asset and returned an empty array when no matching UTxOs were found. For a well-formed asset unit that was never minted, this produced `http/200` with `[]` in the body, whereas the Blockfrost API returns `http/404`.

### Effort

S — Low effort (pure mapping)

### Approach

Look up the asset's `AssetState` up front and return `http/404` when it is absent, before scanning the UTxO set. Add route tests covering the happy path, malformed asset (`http/400`), unknown asset (`http/404`), and state-store failure (`http/500`).

### Why now

1.1425% of observed traffic.

michalrus:
Resolves #1119.

## Context

The `/assets/{asset}/addresses` endpoint scanned the UTxO set for the asset and returned an empty array when no matching UTxOs were found. For a well-formed asset unit that was never minted, this produced `http/200` with `[]` in the body, whereas the Blockfrost API returns `http/404`.

Look up the asset's `AssetState` up front and return `http/404` when it is absent, before scanning the UTxO set. Add route tests covering the happy path, malformed asset (`http/400`), unknown asset (`http/404`), and state-store failure (`http/500`).

This has been tested with `blockfrost-tests`, and the respective test is now passing on Preview:

- https://github.com/blockfrost/blockfrost-tests/blob/594f054ecb5e5a7e68f27abff3d8fe3bab285ea5/src/fixtures/preview/assets/assets-addresses.ts#L28-L35

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved behavior of the asset addresses endpoint for valid, missing, and invalid asset inputs.
  * Added an early missing-asset check to return `NOT_FOUND` more reliably.
  * Enhanced responses when the state lookup fails.

* **Tests**
  * Added integration coverage for `/assets/{asset}/addresses`, including happy path, `BAD_REQUEST`, `NOT_FOUND`, and `INTERNAL_SERVER_ERROR`.

### DIFF
--- a/crates/minibf/src/routes/assets.rs
+++ b/crates/minibf/src/routes/assets.rs
@@ -513,9 +513,16 @@ pub async fn by_subject_addresses<D>(
 ) -> Result<Json<Vec<AssetAddressesInner>>, Error>
 where
     D: Domain + Clone + Send + Sync + 'static,
+    Option<AssetState>: From<D::Entity>,
 {
     let pagination = Pagination::try_from(params)?;
     let asset = hex::decode(&subject).map_err(|_| Error::InvalidAsset)?;
+
+    let entity_key = pallas::crypto::hash::Hasher::<256>::hash(asset.as_slice());
+    domain
+        .read_cardano_entity::<AssetState>(entity_key.as_slice())?
+        .ok_or(StatusCode::NOT_FOUND)?;
+
     let utxoset = domain
         .indexes()
         .utxos_by_asset(&asset)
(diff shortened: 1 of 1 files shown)


crates/minibf/src/routes/assets.rs
```diff
@@ -513,9 +513,16 @@ pub async fn by_subject_addresses<D>(
...
     D: Domain + Clone + Send + Sync + 'static,
+    Option<AssetState>: From<D::Entity>,
 {
     let pagination = Pagination::try_from(params)?;
     let asset = hex::decode(&subject).map_err(|_| Error::InvalidAsset)?;
+
+    let entity_key = pallas::crypto::hash::Hasher::<256>::hash(asset.as_slice());
+    domain
+        .read_cardano_entity::<AssetState>(entity_key.as_slice())?
+        .ok_or(StatusCode::NOT_FOUND)?;
```
scarmuega: just a readability improvement, there's a "entity_exists" that can be used like this.

```rust
if !domain.cardano_entity_exists::<AssetState>(entity_key.as_slice())? {
    return Err(StatusCode::NOT_FOUND.into());
}
```

michalrus: Nice! Fixed in 85badac7948edc0d47debf552d48a243b115d6f6

scarmuega (approved):

PR #1121, opened 2026-07-21
feat(minibf): add `/governance/dreps`

### LINKED ISSUE #1086: minibf: add `/governance/dreps`
Implement the Blockfrost `/governance/dreps` endpoint in the `minibf` crate.

### Effort

S — Low effort (pure mapping)

### Approach

Iterate `DRepState` namespace

Full-range iter over `dreps` ns; sort by `registered_at`

### Why now

0.0232% of observed traffic, used by Lace.

vladimirvolek:
This PR resolves: https://github.com/txpipe/dolos/issues/1086

offcahin metadata cache example in the repo:

image

## Summary by CodeRabbit

* **New Features**
  * Added `GET /governance/dreps` with pagination, ordering, and retired/expired filters.
  * Added DRep lookup at `/governance/dreps/{drep_id}`, supporting legacy, modern, and special identifiers.
  * Added DRep activity, delegation, voting power, and lifecycle details.
  * Added off-chain DRep metadata retrieval with URL validation, size limits, and hash verification.
  * DRep records now retain their first on-chain sighting.
  * Added DRep support to entity update tooling.
* **Documentation**
  * Documented the new governance DReps endpoint.


vladimirvolek:
@copilot resolve the merge conflicts in this pull request

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+    mapping::IntoModel as _,
+    pagination::{Order, Pagination, PaginationParameters},
+    Facade,
+};
+
+/// Stake delegated to each DRep, keyed the same way DRep entities are.
+///
+/// `only` narrows the fold to a single DRep so the by-id route doesn't
+/// materialize the whole map just to read one entry.
+fn drep_stake_map<D: Domain>(
```
slowbackspace: **Severity: 🔴 high**

This fold decodes every `AccountState` on each request, for both routes. On mainnet that is ~1.5M entity decodes per GET, and the by-id route regresses from the previous O(1) read to a full scan (`only` skips map inserts, not the iteration).

The boundary pass already maintains this aggregate: `DRepPowerUpdate` writes the EWRAP `drep_distr` into `DRepState.voting_power`. Since #1228 (merged; not yet in this branch's base, so a rebase is needed anyway — both PRs touch `ewrap/loading.rs`) that value is ledger-exact: it carries the boundary credits (enacted treasury withdrawals, retiring-pool refunds, resolved-proposal deposit gating), verified whole-population-exact against db-sync on preprod/preview and against Koios on mainnet. A live per-account fold bypasses that corrected row and reintroduces the same boundary-timing divergence on the API surface that #1228 just closed in the store.

The semantics differ too. I diffed all 15,047 preview DReps against live BF: 164 amounts differ (live fold vs epoch snapshot). One retired DRep (`drep1y2hlagkcdpc7acrs5d70k6324eha45ntdz4fcwqzu48484gvw3drr`) shows `519640002` here vs `0` on BF, because delegations persist in accounts after retirement while `drep_distr` excludes them.

One caveat on the comparison target: #1228 established that db-sync's own `drep_distr` aggregation disagrees with db-sync's raw tables for some DReps (12 on mainnet at epoch 645), and dolos matches the raw tables exactly. So exact BF `amount` parity is not the goal — serving the ledger-exact `voting_power` is, and small residual diffs vs BF are expected and correct.

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+
+        let stake = account.stake.set().map(|x| x.total()).unwrap_or_default();
+
+        *out.entry(key).or_default() += stake;
+    }
+
+    Ok(out)
+}
+
+fn chain_context<D: Domain>(
```
slowbackspace: **Severity: 🟢 low**

`Facade::get_tip_slot()` already does this in one line with the same 500 mapping (see accounts.rs). It also reads the state cursor rather than the archive tip, so governance routes would agree with the account routes about the current epoch near the tip.

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+        .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?
+        .ok_or(StatusCode::INTERNAL_SERVER_ERROR)?;
+
+    let pparams = domain.get_current_effective_pparams()?;
+
+    Ok((chain, tip, pparams))
+}
+
+pub async fn all_dreps<D: Domain>(
+    Query(params): Query<PaginationParameters>,
```
slowbackspace: **Severity: 🔴 high**

BF supports `retired=`, `expired=`, and `order_by=amount` on this endpoint, and the official test suite covers them (7 matrix cases fail against this branch). Axum silently drops the unknown params, so the divergence is invisible to callers. Verified live: BF preview returns only retired rows for `?retired=true`; dolos returns an unfiltered page.

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+) -> Result<Json<Vec<DrepsInner>>, Error>
+where
+    Option<DRepState>: From<D::Entity>,
+    Option<AccountState>: From<D::Entity>,
+{
+    let pagination = Pagination::try_from(params)?;
+
+    let mut dreps = vec![];
+
+    for item in domain.iter_cardano_entities::<DRepState>(None)? {
```
slowbackspace, line 96: **Severity: 🟡 medium**

This loads and sorts the whole namespace per request and never calls `pagination.enforce_max_scan_limit(domain.config.max_scan_items())` — the cap every other scanning route applies (accounts, addresses, assets, metadata).

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+        let (key, state) = item.map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+
+        let appeared_at = state.first_seen_at.unwrap_or((u64::MAX, usize::MAX));
```
slowbackspace: **Severity: 🟡 medium**

On stores synced before this change every row has `first_seen_at = None`, so the list comes back in entity-key order and never-registered/special DReps are missing entirely (live diff: dolos 15,047 rows vs BF 15,062; the 15 missing are `drep_always_abstain`, `drep_always_no_confidence`, and 13 vote-delegation-only DReps). Full resync fixes it, which is the normal cost for state additions — but worth one line in the CHANGELOG entry so operators expect it.

Small idiom note: sorting by `(first_seen_at.is_none(), first_seen_at, key)` gives the same None-last order without the `u64::MAX` sentinel tuple.

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+    let stake = drep_stake_map(&domain, None)?;
+
+    let states: Vec<_> = dreps
+        .into_iter()
+        .skip(pagination.from())
+        .take(pagination.count)
+        .map(|(_, key, state)| (key, state))
+        .collect();
+
+    let metadata_futures: Vec<_> = states
```
slowbackspace: **Severity: 🟢 low**

The two parallel Vecs stay correct only while they remain index-aligned; a future filter or early-continue in one pass attaches DRep A's metadata to DRep B. One `join_all` over per-item async blocks gives the same concurrency without the positional coupling. That shape would also stop cloning the full `PParamsSet` (cost models included) per row — the builder only reads `drep_inactivity_period()`.

crates/cardano/src/roll/dreps.rs
```diff
@@ -224,10 +224,16 @@ impl BlockVisitor for DRepStateVisitor {
             }
         }
 
+        if let Some(cert) = pallas_extras::cert_as_vote_delegation(cert) {
```
slowbackspace: **Severity: 🔴 high**

Both `DRepSeen` emissions (here and after `cert_drep`) run for certs in phase-2-invalid txs. The committee-cert handling in this same visitor is gated on `tx.is_valid()` for exactly this reason, and db-sync does not apply certs from invalid txs. A collateral-consumed tx would create a phantom DRep row (or stamp a wrong, earlier `first_seen_at`) that BF never records.

crates/minibf/src/routes/governance/dreps.rs
```diff
@@ -0,0 +1,329 @@
...
+        DREP_ALWAYS_NO_CONFIDENCE => Ok(ParsedDRep::special(drep_id, 1)),
+        drep_id => {
+            let (hrp, payload) = bech32::decode(drep_id).map_err(|_| StatusCode::BAD_REQUEST)?;
+
+            match (hrp.as_str(), payload.len()) {
+                ("drep", 29) => {
+                    let header_byte = payload.first().ok_or(StatusCode::BAD_REQUEST)?;
+
+                    // first 4 bits need to be equal to 0010
+                    if header_byte & 0b11110000 != 0b00100000 {
```
slowbackspace: **Severity: 🟡 medium**

CIP-129 defines exactly two valid DRep headers: `0x22` (key hash) and `0x23` (script hash); credential values 0 and 1 are reserved. This check only validates the upper nibble, so `0x20/0x21/0x24…0x2f` pass parsing and fall through to a 404. backend-ryo validates the credential nibble too (`validateDRepId`: `Invalid credential type in header`), and live BF returns 400 for these on both preview and mainnet. Suggest `if header_byte != 0x22 && header_byte != 0x23 { return Err(BAD_REQUEST) }` — `has_script` for CIP-129 ids then also stays trustworthy.

crates/minibf/src/routes/governance/dreps.rs
```diff
@@ -0,0 +1,329 @@
...
+                    }
+
+                    Ok(ParsedDRep::cip129(drep_id, payload))
+                }
+                ("drep", 28) => Ok(ParsedDRep::legacy(
+                    drep_id.to_string(),
+                    payload,
+                    pallas_extras::DREP_KEY_PREFIX,
+                )),
+                ("drep_vkh", 28) => Ok(ParsedDRep::legacy(
```
slowbackspace: **Severity: 🟡 medium**

backend-ryo accepts only the `drep` and `drep_script` prefixes and throws `Invalid drep id prefix` for `drep_vkh`; live BF returns 400 for `drep_vkh1…` on preview and mainnet. Accepting it here is a silent superset — a client that works against dolos breaks on BF. Either drop this arm for parity or document the extension deliberately.

crates/minibf/src/routes/governance/dreps.rs
```diff
@@ -0,0 +1,329 @@
...
+    pub is_legacy: bool,
+    pub state: Option<DRepState>,
+    pub delegated_stake: u64,
+    pub pparams: PParamsSet,
+    pub chain: &'a ChainSummary,
+    pub tip: BlockSlot,
+}
+
+impl<'a> DrepModelBuilder<'a> {
+    fn is_special_case(&self) -> bool {
```
slowbackspace: **Severity: 🟢 low**

`parse_drep_id` already computes `is_special`, and the list path holds the `DRep` enum. Re-deriving specialness from the rendered string couples correctness to the exact output of `bech32_drep`: if the rendering ever changes, the abstain row silently takes the normal path and emits `hex: "00"` plus lifecycle fields from the pseudo-row. An `is_special: bool` field on the builder removes the string round-trip.

crates/minibf/src/routes/governance/dreps.rs
```diff
@@ -0,0 +1,329 @@
...
+pub fn drep_list_item(
+    state: DRepState,
+    delegated_stake: u64,
+    pparams: PParamsSet,
+    chain: &ChainSummary,
+    tip: BlockSlot,
+) -> Result<DrepsInner, StatusCode> {
+    let drep_id = bech32_drep(&state.identifier)?;
+
+    let drep_id_encoded = match &state.identifier {
```
slowbackspace: **Severity: 🟢 low**

This match re-implements `dolos_cardano::model::drep_to_entity_key` byte-for-byte, and the caller already holds the `EntityKey` from iteration (it is the same bytes — it's the key used for the stake map). Passing the key through, or calling the canonical fn, keeps the encoding in one place.

crates/minibf/src/routes/governance/metadata.rs
```diff
@@ -0,0 +1,173 @@
...
+}
+
+fn blocked_url_error(url: &str) -> DrepsInnerMetadataError {
+    DrepsInnerMetadataError::new(
+        MetadataError::ConnectionError,
+        format!("Error Offchain Drep: Refused to fetch metadata from {url}, only http and https URLs are allowed."),
+    )
+}
+
+fn is_fetchable(url: &str) -> bool {
```
slowbackspace, line 94: **Severity: 🔴 high**

Scheme-only validation means the node fetches whatever URL a DRep registered on-chain: `http://169.254.169.254/…` or `http://localhost:PORT/…` are fetched from the node's network position on every list request (SSRF), and redirects are followed. Combined with no caching, a page render fires up to `count=100` outbound GETs with 5s timeouts, repaid on every poll. backend-ryo never fetches live — it serves db-sync's offline-fetched `off_chain_vote_data`. Suggest caching by `(url, hash)` at minimum (content is hash-pinned, so a hit never revalidates) and blocking private/link-local ranges. Also worth unifying with the pool metadata fetcher in pools.rs, which has the same flow minus the size cap this version adds.

crates/minibf/src/routes/governance/metadata.rs
```diff
@@ -0,0 +1,173 @@
...
+    if !is_fetchable(&anchor.url) {
+        return errored(out, blocked_url_error(&anchor.url));
+    }
+
+    let mut response = match client.get(&anchor.url).send().await {
+        Ok(response) => response,
+        Err(_) => return errored(out, connection_error(&anchor.url)),
+    };
+
+    if response.status() != StatusCode::OK {
```
slowbackspace: **Severity: 🟢 low**

Minor: any 2xx other than 200 (e.g. 203/206 via a proxy) becomes `HTTP_RESPONSE_ERROR` here. Also `http_client()` caches a failed `Client::build()` forever, turning one TLS-init failure into permanent `CONNECTION_ERROR`s.

crates/minibf/src/routes/governance/metadata.rs
```diff
@@ -0,0 +1,173 @@
...
+            out,
+            hash_mismatch_error(
+                &anchor.url,
+                anchor.content_hash.as_slice(),
+                actual_hash.as_ref(),
+            ),
+        );
+    }
+
+    out.json_metadata = serde_json::from_slice(&body).ok();
```
slowbackspace: **Severity: 🟡 medium**

`.ok()` swallows invalid JSON. The spec says failed validation must populate `error` with `json_metadata`/`bytes` null, and both the openapi enum and backend-ryo have a `DECODE_ERROR` code for exactly this case. As written, hash-matching non-JSON bytes produce `json_metadata: null` + populated `bytes` + no error — a shape BF clients never see.

crates/minibf/src/test_support.rs
```diff
@@ -231,3 +255,44 @@ impl TestApp {
...
+        self.cfg = cfg;
+        self
+    }
+
+    pub fn with_protocol(mut self, protocol: usize) -> Self {
+        self.force_protocol = Some(protocol);
+        self
+    }
+
+    pub fn build(self) -> TestApp {
```
slowbackspace, line 287: **Severity: 🟢 low**

This re-implements `TestApp::from_domain` (same module) verbatim with `TestFault::None` hardcoded. Ending with `TestApp::from_domain(domain, vectors, None)` keeps one copy of the app wiring and gives the builder fault support for free later.

slowbackspace (commented):
I tested this PR against live Blockfrost (preview and mainnet) and the official blockfrost-tests suite. Summary below; details in the inline comments.

Each inline comment carries a severity label: 🔴 high = fix before merge, 🟡 medium = BF-parity or spec divergence, 🟢 low = cleanup/nit.

## What works well

- All 5 official by-id fixtures pass (CIP-129, legacy, script-hash, special IDs). All pagination-error cases pass.
- I diffed the full DRep population against live BF preview: 15,047 shared rows. `retired`, `expired`, `last_active_epoch`, and `has_script` show **zero mismatches**. The expiry math (including the `<` boundary fix) is correct.
- `DRepSeen` matches BF semantics. BF lists never-registered DReps (vote-delegation targets, `active_epoch: null`) and the special DReps. Materializing them is right.

## Divergences found (live-verified)

| Case | dolos | BF (preview + mainnet) |
|---|---|---|
| CIP-129 header `0x20/0x21/0x24` | 404 | 400 |
| `drep_vkh1…` id | 200 | 400 |
| `?retired=` / `?expired=` / `?order_by=amount` | ignored | filtered/sorted |
| `amount` | live account fold | epoch-boundary `drep_distr` snapshot (164/15047 rows differ; one retired DRep: dolos `519640002`, BF `0`) |


7 of the official `governance/dreps` list-matrix tests fail on the missing query params.

## Main concerns

1. Every request to both routes scans the full accounts namespace. The by-id route regresses from an O(1) `voting_power` read to O(all accounts). See inline comment.
2. The list route live-fetches up to 100 attacker-controlled anchor URLs per request, uncached, with no private-IP guard (SSRF). backend-ryo serves this from db-sync's offchain store instead.
3. `DRepSeen` fires for certs in phase-2-invalid txs. Sibling cert handlers guard on `tx.is_valid()`.

slowbackspace (changes requested):

vladimirvolek:
@copilot resolve the merge conflicts in this pull request

vladimirvolek:
@copilot resolve the merge conflicts in this pull request

PR #1123, opened 2026-07-22
feat(minibf): add block txs CBOR endpoints

### LINKED ISSUE #1069: minibf: add `/blocks/{hash_or_number}/txs/cbor`
Implement the Blockfrost `/blocks/<block>/txs/cbor` endpoint in the `minibf` crate.

### Effort

S — Low effort (pure mapping)

### Approach

Block txs \+ re-encode each tx

### Why now

0.6110% of observed traffic.

### LINKED ISSUE #1116: minibf: add `/blocks/latest/txs/cbor`
Implement the Blockfrost `/blocks/latest/txs/cbor` endpoint in the `minibf` crate.

### Effort

S — Low effort (pure mapping)

### Approach

Same, for tip

### Why now

0.0001% of observed traffic.

slowbackspace:
## Summary

Implements the Blockfrost-compatible block transaction CBOR endpoints:

- `GET /blocks/{hash_or_number}/txs/cbor`
- `GET /blocks/latest/txs/cbor`

Both return `[{ tx_hash, cbor }]` per transaction in the block, supporting the standard `count`/`page`/`order` pagination parameters.

## Implementation

- New `IntoModel<Vec<BlockContentTxsCborInner>>` impl on `BlockModelBuilder` — the CBOR is emitted from the transaction bytes preserved in the raw block body (Pallas `KeepRaw`): `MultiEraTx::encode` re-serializes only the thin outer envelope (array header + validity flag), while body, witness set, and auxiliary data are written verbatim from the preserved raw slices — so the output is byte-identical to the on-wire encoding (verified against live Blockfrost).
- Handlers mirror the existing `by_hash_or_number_txs` / `latest_txs` shape, including 400-vs-404 semantics (`Error::InvalidBlockHash`/`InvalidBlockNumber` for malformed input, 404 for well-formed but absent blocks).
- Docs endpoint table updated in `docs/content/apis/minibf.mdx`.

## Testing

- 9 new unit tests following the crate's standard matrix: happy path (with CBOR round-trip decode asserting hash consistency), order asc/desc, pagination, bad request, not found, and fault-injected internal error.
- Verified against **live Blockfrost preview API**: responses are byte-identical for blocks by hash and by number, with `order=desc`, with `count`/`page` pagination, and for the 400/404 error bodies.
- `cargo +nightly fmt --check`, `cargo clippy --workspace --all-targets --all-features` (zero warnings), `cargo test -p dolos-minibf` all green after rebasing onto v1.5.0.

resolves https://github.com/txpipe/dolos/issues/1116
resolves https://github.com/txpipe/dolos/issues/1069

## Summary by CodeRabbit

* **New Features**
  * Added endpoints to retrieve transactions from the latest or a specified block with CBOR-encoded data.
  * Added pagination and ascending/descending ordering support for CBOR transaction results.
  * Responses include transaction hashes and encoded transaction payloads.

* **Documentation**
  * Updated the API documentation with the new CBOR transaction endpoints.

### DIFF
--- a/crates/minibf/src/mapping.rs
+++ b/crates/minibf/src/mapping.rs
@@ -33,6 +33,7 @@ use blockfrost_openapi::models::{
     block_content::BlockContent,
     block_content_addresses_inner::BlockContentAddressesInner,
     block_content_addresses_inner_transactions_inner::BlockContentAddressesInnerTransactionsInner,
+    block_content_txs_cbor_inner::BlockContentTxsCborInner,
     tx_content::TxContent,
     tx_content_cbor::TxContentCbor,
     tx_content_delegations_inner::TxContentDelegationsInner,
@@ -2264,6 +2265,25 @@ impl<'a> IntoModel<Vec<String>> for BlockModelBuilder<'a> {
             .map(|tx| tx.hash().to_string())
             //.sorted()
             //.map(|tx| BlockContentAddressesInnerTransactionsInner { tx_hash: tx })
+            .collect();
+
+        Ok(txs)
+    }
+}
+
+impl<'a> IntoModel<Vec<BlockContentTxsCborInner>> for BlockModelBuilder<'a> {
+    type SortKey = ();
+
+    fn into_model(self) -> Result<Vec<BlockContentTxsCborInner>, StatusCode> {
+        let block = &self.block;
+
+        let txs = block
+            .txs()
+            .iter()
+            .map(|tx| BlockContentTxsCborInner {
+                tx_hash: tx.hash().to_string(),
+                cbor: hex::encode(tx.encode()),
+            })
             .collect();
 
         Ok(txs)
--- a/crates/minibf/src/routes/blocks.rs
+++ b/crates/minibf/src/routes/blocks.rs
@@ -786,6 +786,23 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn blocks_latest_txs_cbor_paginated() {
+        let app = TestApp::new();
+        let block = app.vectors().blocks.last().expect("missing block vectors");
+        let (status, bytes) = app
+            .get_bytes("/blocks/latest/txs/cbor?count=1&page=2")
+            .await;
+        assert_eq!(status, StatusCode::OK);
+
+        let txs: Vec<BlockContentTxsCborInner> =
+            serde_json::from_slice(&bytes).expect("failed to parse paginated latest txs cbor");
+
+        let expected: Vec<String> = block.tx_hashes.iter().skip(1).take(1).cloned().collect();
+        let hashes: Vec<String> = txs.iter().map(|tx| tx.tx_hash.clone()).collect();
+        assert_eq!(hashes, expected);
+    }
+
+    #[tokio::test]
     async fn blocks_latest_txs_cbor_internal_error() {
         let app = TestApp::new_with_fault(Some(TestFault::ArchiveStoreError));
         assert_status(
--- a/crates/minibf/src/routes/blocks.rs
+++ b/crates/minibf/src/routes/blocks.rs
@@ -3,7 +3,9 @@ use axum::{
     http::StatusCode,
     Json,
 };
-use blockfrost_openapi::models::block_content::BlockContent;
+use blockfrost_openapi::models::{
+    block_content::BlockContent, block_content_txs_cbor_inner::BlockContentTxsCborInner,
+};
 use dolos_cardano::ChainSummary;
 use dolos_core::{archive::Skippable as _, ArchiveStore as _, BlockBody, Domain};
 use futures::future::try_join_all;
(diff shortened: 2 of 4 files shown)


crates/minibf/src/routes/blocks.rs
```diff
@@ -595,4 +741,58 @@ mod tests {
...
+        for tx in txs {
+            let cbor = hex::decode(&tx.cbor).expect("cbor is not valid hex");
+            let decoded = pallas::ledger::traverse::MultiEraTx::decode(&cbor)
+                .expect("cbor is not a decodable tx");
+            assert_eq!(decoded.hash().to_string(), tx.tx_hash);
+        }
+    }
+
+    #[tokio::test]
+    async fn blocks_latest_txs_cbor_order_desc() {
```
michalrus, line 773: A little asymmetry -- `blocks_latest` lacks the `_paginated` test

crates/minibf/src/mapping.rs
```diff
@@ -2270,6 +2271,25 @@ impl<'a> IntoModel<Vec<String>> for BlockModelBuilder<'a> {
...
+
+    fn into_model(self) -> Result<Vec<BlockContentTxsCborInner>, StatusCode> {
+        let block = &self.block;
+
+        let txs = block
+            .txs()
+            .iter()
+            .map(|tx| BlockContentTxsCborInner {
+                tx_hash: tx.hash().to_string(),
+                cbor: hex::encode(tx.encode()),
```
michalrus, line 2285: The PR description says "no re-encoding takes place", but `tx` is a `&MultiEraTx<'_>`, and in the definition of its `encode()` I see several `minicbor::*` calls. 🤔

michalrus: Also we're `hex::encode`’ing all elements even before pagination's `.skip(…).take(…)` above, but that's probably cheap enough?

michalrus (commented):
I have tested this with `blockfrost-tests` on Preview, and they pass:
- https://github.com/blockfrost/blockfrost-tests/blob/f999931323e74d5dd443a97e88b0e416e2d10b7c/src/fixtures/common/blocks/latest/txs/cbor.ts
- https://github.com/blockfrost/blockfrost-tests/blob/f999931323e74d5dd443a97e88b0e416e2d10b7c/src/fixtures/preview/blocks/hash-or-number-txs-cbor.ts

A couple of minor comments below:

crates/minibf/src/routes/blocks.rs
slowbackspace: Good catch — added `blocks_latest_txs_cbor_paginated` in 91d079b3.

crates/minibf/src/mapping.rs
slowbackspace: You're right — "no re-encoding" was imprecise, reworded.

slowbackspace: Yeah — deliberately kept for symmetry with the sibling `/txs` handlers, which also materialize the full list before `skip`/`take`. Worst case is a few hundred txs at a few KB each, so the wasted hex is on the order of a megabyte of throwaway allocation per request — micro/milliseconds next to the archive read. If it ever shows up in a profile the fix is to slice the tx list before mapping, but that felt like premature restructuring here.

michalrus (approved):
LGTM!

michalrus:
@coderabbitai, re: your edit of the PR description, I don't understand how this is a bug fix specifically? Just the classification.

> - **Bug Fixes**
>
>   - Added validation for invalid block references, missing blocks, and archive retrieval failures.

I mean it has never been a bug…?

michalrus:
@coderabbitai,

> The Bug Fixes section should be removed

Why do you say it "should be removed" if it's _your_ change, in a section owned by _you_, under `<!-- This is an auto-generated comment: release notes by coderabbit.ai -->`? Just do it maybe…?

slowbackspace:
@coderabbitai just fix your summary by removing bug fixes section. This PR introduces new features, it doesn't fix anything.

crates/minibf/src/routes/blocks.rs
```diff
@@ -401,6 +403,34 @@ where
...
+    let hash_or_number = parse_hash_or_number(&hash_or_number)?;
+    let block = load_block_by_hash_or_number(&domain, &hash_or_number).await?;
+
+    let model = BlockModelBuilder::new(&block)?;
+
+    let txs: Vec<BlockContentTxsCborInner> = model.into_model()?;
+    let txs = match pagination.order {
+        Order::Asc => txs,
+        Order::Desc => txs.into_iter().rev().collect(),
+    };
```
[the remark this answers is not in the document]
slowbackspace: Deliberate: this mirrors the adjacent handlers (`by_hash_or_number_txs`, `latest_txs`, `previous`/`next`) which all reverse the same way — the pattern appears 7× in this file, 5 of them pre-existing. The cost is one pointer-level allocation per request (the item structs move; the CBOR payload strings don't), negligible next to the encoding work above it. Happy to do a file-wide `Vec::reverse()` sweep as a separate style PR if maintainers want the idiom changed everywhere.

scarmuega (approved):

PR #1124, opened 2026-07-22
fix(minibf): return affected addresses from /blocks/{hash_or_number}/addresses

### LINKED ISSUE #1068: minibf: fix `/blocks/{hash_or_number}/addresses`
Fix implementation of the Blockfrost `/blocks/<block>/addresses` endpoint in the `minibf` crate.

Bug: wrong shape vs real BF. Route exists (`crates/minibf/src/routes/blocks.rs:404`) but returns tx-hash list (`BlockModelBuilder::into_model`), not affected addresses.

### Effort

M — Medium (data exists, needs derivation)

### Approach

Currently wrong shape (tx-hash list). Real impl = decode block + resolve input addresses.

### Why now

1.7205% of observed traffic.

slowbackspace:
Closes #1068

## Problem

`/blocks/{hash_or_number}/addresses` returned a plain list of tx hashes — the same shape as `/blocks/{hash_or_number}/txs` — instead of Blockfrost's contract: one entry per affected address with the transactions that affected it. The existing `IntoModel<Vec<BlockContentAddressesInner>>` impl was dead code, and it only looked at outputs, emitted one entry per output (no grouping), and `unwrap()`ed address decoding.

## Implementation

An address is affected by a tx if it received a regular output or had a UTxO spent by a regular input. Since inputs are only `(tx_hash, index)` pointers, input-side addresses require resolving the referenced transactions — done with the same dep-resolution pattern `TxModelBuilder` uses for `/txs/{hash}/utxos`: `BlockModelBuilder::required_input_deps()` collects the deduped source-tx hashes across the block, the handler fetches them via `get_tx_batch`, and `load_dep` feeds them back for resolution during mapping.

Response semantics (matching Blockfrost):
- one entry per address, sorted bytewise-ascending by encoded address string; the `order` param is ignored, as in Blockfrost
- per address, tx hashes in block order, each at most once regardless of how many outputs/inputs touched it
- `count`/`page` paginate whole address entries; past-the-end pages return `200 []`

## Deliberate divergences from Blockfrost

- **Phase-2-failed txs are included**: their collateral inputs and collateral-return outputs move funds, so those addresses are affected in ledger terms. Blockfrost omits them only because its SQL reads db-sync's `tx_in`/`tx_out` (valid txs only) — treated as a Blockfrost bug rather than behavior to reproduce. Documented in a code comment.
- **Genesis virtual block**: Blockfrost returns `200 []` for `/blocks/0/addresses` / the genesis hash; dolos returns 404 (the genesis block isn't in the archive). Kept simple deliberately — no practical traffic on that path.

## Testing

Verified locally against the official [blockfrost-tests](https://github.com/blockfrost/blockfrost-tests) suite targeting a dolos preview node bootstrapped from a full snapshot and synced to tip:

```
SERVICE_NAME=blockfrost-tests yarn test:preview -t "blocks/:hash_or_number/addresses"
```

Result: **4/4 passed** — `blocks/d452532f…/addresses`, `blocks/1846333/addresses`, and both `?count=3&page=1` variants, matching the pinned fixture responses (which include input-side addresses on a deep-history block).

- 7 new unit tests: happy path (grouping/dedup + alphabetical ordering), `order` param ignored, pagination, bad request, not found, fault-injected 500.
- Live verification against Blockfrost preview: **28/28 blocks byte-identical** (including empty blocks and a block with intra-block tx chaining), plus pagination windows, past-the-end pages, by-hash lookups, and 400/404 error bodies.
- `cargo +nightly fmt --check`, `cargo clippy --workspace --all-targets --all-features` (zero warnings), `cargo test -p dolos-minibf` (227 passing).

## Summary by CodeRabbit

- **New Features**
  - Block address endpoints now return, for each address, the set of transaction hashes that “touch” it, combining both produced outputs and effects from consumed inputs.

- **Bug Fixes**
  - Address ordering is now deterministic (alphabetical) and consistently ignores the `order` parameter.
  - Missing archive dependencies are handled gracefully to keep address mappings correct, including collateral-return resolution.

- **Tests**
  - Expanded `/blocks/{hash}/addresses` coverage for alphabetical sorting, `count`/`page` pagination (including past-the-end returning empty), and 400/404/500 error handling, plus an updated collateral-return regression test.

### DIFF
--- a/crates/minibf/src/routes/blocks.rs
+++ b/crates/minibf/src/routes/blocks.rs
@@ -449,6 +449,9 @@ where
     let deps = builder.required_input_deps();
     let deps = domain.get_tx_batch(deps).await?;
 
+    // deps missing from the archive (possible on nodes without full history)
+    // are skipped, omitting their addresses from the response — the same
+    // graceful degradation as /txs/{hash}/utxos
     for (key, cbor) in deps.iter() {
         if let Some(cbor) = cbor {
             builder.load_dep(*key, cbor)?;
--- a/crates/minibf/src/routes/blocks.rs
+++ b/crates/minibf/src/routes/blocks.rs
@@ -909,6 +909,84 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn blocks_by_hash_or_number_addresses_past_the_end_page() {
+        let app = TestApp::new();
+        let block = app.vectors().blocks.first().expect("missing block vectors");
+
+        let path = format!("/blocks/{}/addresses?count=100&page=99", block.block_hash);
+        let addresses = get_addresses(&app, &path).await;
+
+        assert!(
+            addresses.is_empty(),
+            "past-the-end page must be an empty 200"
+        );
+    }
+
+    /// Exercises the `produces_at` edge for input-side address resolution.
+    /// Invalid script txs don't produce their regular outputs, but they can
+    /// produce a collateral return at the next output index. When a later tx
+    /// spends that collateral return, block-address mapping must resolve the
+    /// collateral-return address from the dependency tx CBOR.
+    #[test]
+    fn block_addresses_resolves_spent_collateral_return_address() {
+        use dolos_core::EraCbor;
+        use dolos_testing::synthetic::{build_synthetic_blocks, SyntheticBlockConfig};
+        use pallas::ledger::traverse::{Era, MultiEraBlock, MultiEraTx};
+
+        // Invalid Conway tx with no regular outputs and one collateral-return
+        // output at index 0.
+        let dep_tx_cbor = hex::decode(
+            "84a600d9010281825820010101010101010101010101010101010101010101010101010101010101010100018002070dd901028182582002020202020202020202020202020202020202020202020202020202020202020010a200581d607393621349f3555f70c975392c84879ba400d08d14ab3d572d7ece80011a0016e360111a0007a120a0f4f6",
+        )
+        .expect("invalid dep tx fixture hex");
+        let collateral_return_index = 0usize;
+
+        let dep_tx =
+            MultiEraTx::decode_for_era(Era::Conway, &dep_tx_cbor).expect("failed to decode dep tx");
+        assert!(
+            dep_tx.output_at(collateral_return_index).is_none(),
+            "regular outputs do not include collateral return index"
+        );
+        let collateral_return_address = dep_tx
+            .produces_at(collateral_return_index)
+            .expect("collateral return must be produced at index 0")
+            .address()
+            .expect("collateral return must have an address")
+            .to_string();
+
+        let (blocks, _, _) = build_synthetic_blocks(SyntheticBlockConfig::default());
+        let raw = blocks.first().expect("missing synthetic block");
+
+        let block = MultiEraBlock::decode(raw).expect("failed to decode block");
+        let txs = block.txs();
+        let spender = txs.get(1).expect("fixture needs a second tx");
+        let consumed = spender.consumes();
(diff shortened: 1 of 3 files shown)


slowbackspace:
@coderabbitai your summary is wrong. Bug fixes section includes
"Improved handling of invalid block identifiers, missing blocks, and archive retrieval errors.", but this PR doesn't address any of that. Check the original PR description and improve your summary.

slowbackspace:
@coderabbitai On the silently-skipped missing deps: this is deliberate, and consistent with the crate's existing behavior — `/txs/{hash}/utxos` degrades exactly the same way when a spent tx is outside the archive (unresolved inputs are served with missing fields, no error, no log). On a full-archive node — the deployment profile that actually matches Blockfrost's semantics — every dep resolves and the gap doesn't exist; the fixture runs and live byte-parity sweeps confirm that.

Making one endpoint strict (or noisy) while its siblings degrade silently would trade a consistency problem for a visibility one. The right fix is crate-wide — e.g. a single policy at the `get_tx_batch`/facade level (log, or a strict mode that 4xx/5xxs when history is insufficient) applied to all history-dependent endpoints. That deserves its own issue rather than a one-endpoint deviation in this PR. Added a code comment documenting the degradation in e1b8cc3c so the behavior reads as a decision rather than an oversight.

crates/minibf/src/routes/blocks.rs
```diff
@@ -405,24 +407,35 @@ pub async fn by_hash_or_number_addresses<D>(
...
-    let model = BlockModelBuilder::new(&block)?;
+    let mut builder = BlockModelBuilder::new(&block)?;
 
-    let txs: Vec<String> = model.into_model()?;
-    let txs = match pagination.order {
-        Order::Asc => txs,
-        Order::Desc => txs.into_iter().rev().collect(),
-    };
+    let deps = builder.required_input_deps();
+    let deps = domain.get_tx_batch(deps).await?;
```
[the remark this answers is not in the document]
slowbackspace: `get_tx_batch` is a pre-existing shared helper (also behind `/txs/{hash}/utxos` and `/txs/{hash}/redeemers`) — this PR only adds a caller. The lookups are local embedded-DB reads, not remote calls; measured request times for this endpoint sit at ~30–60 ms on real preview blocks. Parallelizing it would change all callers, so it belongs in its own PR.

crates/minibf/src/routes/blocks.rs
```diff
@@ -812,4 +824,190 @@ mod tests {
...
+    fn block_addresses_resolves_spent_collateral_return_address() {
+        use dolos_core::EraCbor;
+        use dolos_testing::synthetic::{build_synthetic_blocks, SyntheticBlockConfig};
+        use pallas::ledger::traverse::{Era, MultiEraBlock, MultiEraTx};
+
+        // Invalid Conway tx with no regular outputs and one collateral-return
+        // output at index 0.
+        let dep_tx_cbor = hex::decode(
+            "84a600d9010281825820010101010101010101010101010101010101010101010101010101010101010100018002070dd901028182582002020202020202020202020202020202020202020202020202020202020202020010a200581d607393621349f3555f70c975392c84879ba400d08d14ab3d572d7ece80011a0016e360111a0007a120a0f4f6",
+        )
```
scarmuega: I think this would be more useful as part of the synthetic data module in the testing crate, so that we can inject and test faulty txs in different scenarios. Having it inlined here for this particular endpoint will invite duplication.

slowbackspace: addressed in 4fe22cb8e916284bc149981892aa0cf548eedf3f

PR #1163, opened 2026-08-04
perf(storage): hold exact record keys inline

scarmuega:
**Plan:** `dolos-stelae-store-apis-review-cleanups`, item 5.

**Done criterion (item 5):** `ExactRecord` construction and `build_exact_key` allocate nothing per record; `test_prehashed_key_matches_delta_key` still passes. **Met.**

## What changed

An exact key is fixed-width per kind — 32 bytes for a hash, 8 for a block number — so `ExactRecord.key: Vec<u8>` bought a heap allocation and a pointer chase for bytes that fit in the record. A bulk export makes one record *per entry in the store*, and `build_exact_key` allocated a `Vec` per write and per point query that fjall's batch copied out of and freed immediately.

The key is now `[u8; MAX_EXACT_KEY_LEN]` trimmed by `ExactRecord::key()`, which makes the record `Copy`. `build_exact_key` writes into a stack buffer.

**Ordering is unaffected.** Within a kind every key is the same width, so the zero padding is identical and is never reached before `(kind, key)` has decided. Pinned by `exact_record_ord_is_kind_then_key`. `Debug` is hand-written so the padding does not show up in test output.

## The part worth more than the allocation

Making the field private means `ExactRecord::new` is the only way in — so it becomes fallible, and becomes the **single width-validation site**, replacing the duplicated checks in `exact::insert_prehashed` (fjall) and `MemoryIndexWriter::append_prehashed`.

The review's restore-choke-point guarantee gets *stronger*, not weaker: a wrong-width record can no longer be constructed at all, so the write paths have nothing left to check. The fjall record iterator picks the check up for on-disk entries, where a wrong width is a malformed entry and therefore terminal, per the traversal's error policy.

`build_exact_key` returns `None` for key data wider than any kind stores. No such entry can exist, so a query for one misses rather than erroring — which is exactly what the `Vec` version did, by building a key nothing matched.

The memory store's exact map keys on the same inline form, or `iter_exact_records` would have kept cloning per record.

## Verification

Byte-identical on disk and on the wire — nothing about the stored key layout moved:

```
cargo test --workspace --all-targets      # 24 test binaries, all ok
cargo test --test index_roundtrip         # byte-identity roundtrip, both live backends
cargo clippy --all-targets --all-features -- -D warnings
cargo +nightly fmt --all -- --check
```

`test_prehashed_key_matches_delta_key` is unchanged in substance and passes. New tests: `max_exact_key_len_covers_every_kind` (which also asserts the constant is the widest kind rather than merely an upper bound), `exact_record_rejects_a_wrong_width_key`, `exact_record_ord_is_kind_then_key`, and `test_exact_key_wider_than_any_kind_has_no_key`.

## Semver note

The plan flagged `ExactRecord`'s public shape as semver-relevant. It is not, in practice: the workspace does not publish (`[workspace.metadata.release] publish = false`), and `ExactRecord` has five construction sites, all in-tree.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of exact-index keys with consistent length validation.
  * Invalid or incorrectly sized keys are now rejected safely during updates.
  * Malformed lookup keys now return no match instead of causing errors.
  * Improved exact-key reconstruction during iteration and query processing.
  * Added safeguards to prevent invalid keys from being stored or matching existing records.

### DIFF
--- a/crates/core/src/builtin/memory/index.rs
+++ b/crates/core/src/builtin/memory/index.rs
@@ -32,12 +32,28 @@ use std::sync::{Arc, Mutex, RwLock};
 
 use crate::indexes::{
     key_hash, ArchiveIndexDelta, ExactKind, ExactRecord, IndexDelta, IndexError, IndexRecord,
-    IndexStore, IndexWriter, KeyHash, TagDimension, TagRecord,
+    IndexStore, IndexWriter, KeyHash, TagDimension, TagRecord, MAX_EXACT_KEY_LEN,
 };
 use crate::{BlockSlot, ChainPoint, TxoRef, UtxoSet};
 
 /// A stored archive tag: the traversal contract's sort key, used as the key.
 type ArchiveTag = (Cow<'static, str>, KeyHash, BlockSlot);
+
+/// An exact key, zero-padded to the widest kind — the same inline shape
+/// [`ExactRecord`] carries, so a record goes in and comes back out without a
+/// heap allocation on either side.
+type ExactKey = [u8; MAX_EXACT_KEY_LEN];
+
+/// Pad a fixed-width key into the map's key type.
+///
+/// The width itself is [`ExactRecord::new`]'s to check; by the time a key
+/// reaches here it is either a record's own or built from a typed delta.
+fn exact_key(key: &[u8]) -> ExactKey {
+    let mut buf = [0u8; MAX_EXACT_KEY_LEN];
+    let len = key.len().min(MAX_EXACT_KEY_LEN);
+    buf[..len].copy_from_slice(&key[..len]);
+    buf
+}
 
 /// The store's whole contents, guarded as one unit so a commit is atomic.
 #[derive(Default, Clone)]
@@ -48,7 +64,9 @@ struct Tables {
     /// store, and the set is mutable rather than append-only.
     utxo_tags: BTreeMap<(TagDimension, Vec<u8>), HashSet<TxoRef>>,
     archive_tags: BTreeSet<ArchiveTag>,
-    exact: BTreeMap<(ExactKind, Vec<u8>), BlockSlot>,
+    /// Keyed on the record's own inline key rather than a `Vec`, so
+    /// `iter_exact_records` copies rather than allocates per record.
+    exact: BTreeMap<(ExactKind, ExactKey), BlockSlot>,
 }
 
 /// A single mutation, recorded by a writer and replayed at commit. See the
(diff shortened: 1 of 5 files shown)


scarmuega:
Reviewed the CodeRabbit finding. **It's correct, and it was a defect this PR introduced.** Fixed in 8c6604ac.

## Confirming it first

Before changing anything I wrote the reported case as a test against this branch:

```rust
// store a malformed 33-byte block hash, then query the well-formed
// 32-byte hash that happens to be its prefix
assert_eq!(memory.slot_by_block_hash(&vec![0xAB; 32]).unwrap(), None);
```
```
assertion `left == right` failed: querying a 32-byte hash returned the slot of a different, 33-byte one
  left: Some(100)
 right: None
```

So a *well-formed* lookup returned the wrong block's slot. On `main` this could not happen: the memory store kept the 33-byte key verbatim, and a 32-byte query missed. The inline buffer I added is what introduced the aliasing — padding or truncating to fit makes a key collide with the key it was adjusted into.

The second half of the report is right too. fjall refused keys over 32 bytes but happily persisted *short* ones, and `ExactRecord::new` then refuses to decode them during traversal — so a single short hash makes the whole store un-exportable, since the iterator's error policy is terminal.

## The fix

Both backends now check against `ExactKind::key_len()` at the delta path, which is the one place the width is not already type-enforced (`ArchiveIndexDelta` carries `Vec<u8>` hashes).

- **Write** — a wrong-width hash fails the batch rather than being dropped. A block whose hash did not land is a block no hash lookup can reach, and an index that quietly omits it looks complete. This also matches what `ExactRecord::new` and `insert_prehashed` already did.
- **Query** — a wrong-width key is a miss, not an error: a key that could not have been stored cannot be found.

`build_exact_key` now takes an `ExactKind` instead of a dimension string. That drops `test_any_dimension_works`, which was asserting a chain-agnosticism the exact keyspace does not actually have — unlike tag dimensions, its set is closed by `ExactKind`. Replaced by `test_only_the_kind_width_has_a_key`, which walks every kind and checks width, width−1, width+1 and empty.

## Regression tests

`malformed_exact_keys_are_refused` and `malformed_exact_queries_miss`, both in the cross-backend conformance suite so they run against fjall *and* the memory store. Verified against the pre-fix code — both fail, the second with `left: Some(432000), right: None`, which is the prefix collision.

Full gate re-run: 24 test binaries green, clippy `-D warnings`, `fmt --check`.

One correction to my own PR description while I'm here: it said the change was byte-identical with no behavioural change. That holds for well-formed input — every real Cardano block — but not for malformed input, where behaviour has now deliberately changed in both backends, and where before this fix it had changed *accidentally*.

PR #1184, opened 2026-08-08
feat(stelae): authenticate a registry with basic credentials

scarmuega:
## Plan

`plans/dolos-stelae-registry-basic-auth.md` — *Stelae: Basic credentials for the registry client* (Brain/txpipe, `org/coder`). The Cloudflare registry deployment plan awaits it: its validation gates run through this client.

## What changed

**Client surface.** `stelae::oci::Options` gains an `auth` value — anonymous / bearer / basic — supplied by the caller. The crate keeps the environment grammar (it owns the variable names) in `Auth::from_env`, but nothing calls it behind a caller's back: which identity a program authenticates as is that program's policy. `Auth`'s `Debug` redacts, so a password cannot reach a backtrace through `Options`.

**Consumer surface.** `[stelae.registry]` in `dolos.toml` names the `user`; the password is compiled in. `dolos init` seeds the section from `dolos_core::config::OFFICIAL_REGISTRY_USER` and writes no password, and `StelaeRegistryConfig::password()` falls back to `OFFICIAL_REGISTRY_PASSWORD` when the file names none. A password copied into every generated config has to be found again in every one of them; a compiled-in default rotates with a release. A node pointing at a private registry sets `password` and keeps its own.

Both constants ship empty — one site, two lines, filled once the deployment plan provisions the registry. Until then `init` writes no section rather than one naming nobody.

**Publisher surface.** `STELAE_REGISTRY_USER` / `STELAE_REGISTRY_PASSWORD`, read where `STELAE_REGISTRY_TOKEN` is. `dolos_snapshot::registry::auth` resolves both directions in one rule: environment wins, configuration is the fallback, neither is anonymous. A token and a pair together is a refusal naming both; so is half a pair.

**Testing.** The canonical e2e harnesses are rebound, not duplicated: every registry they spawn is behind htpasswd and every transport carries the pair. Both server families are configured unconditionally — `distribution` reads `REGISTRY_AUTH_*` from the environment, `zot` reads `/etc/zot/config.json`, and neither notices the other's — so there is no per-image branch. `credentials_are_required` (stelae) and `a_node_authenticates_with_its_configured_pair` (dolos-snapshot) assert the server really refuses an unauthenticated request; that is also the fixture's honesty check, since a server it could not authenticate would run anonymous with every other test passing. A 401 read as absence would silently restart a publisher's history chain, so that refusal is now checked against a server that really answers 401.

## Done criterion

**1. Config credentials on restore, env pair on publish, precedence, and the refusal.** Against a real htpasswd `registry:2`, with `[stelae.registry]` deliberately holding the wrong password:

```
############ 1. config credentials only (deliberately wrong)
Error:   × restoring the stele
  ╰─▶ Not authorized: url http://127.0.0.1:57120/v2/dolos/mainnet/manifests/latest

############ 2. same config, correct pair exported
  ╰─▶ Registry error: url http://127.0.0.1:57120/v2/dolos/mainnet/manifests/
      latest, envelope: OCI API errors: [OCI API error: manifest unknown]
```

The environment got past auth; the only remaining complaint is that the repository is empty. And the refusal:

```
$ STELAE_REGISTRY_TOKEN=… STELAE_REGISTRY_USER=… STELAE_REGISTRY_PASSWORD=… \
    dolos bootstrap stelae --source oci://…
Error:   × opening the repository
  ╰─▶ STELAE_REGISTRY_TOKEN and STELAE_REGISTRY_USER/STELAE_REGISTRY_PASSWORD
      are both set; registry credentials come from one of the two and which
      one was meant is not something to guess at — unset the one you did not
      mean
```

**2. A fresh `dolos init` seeds the section, without a password.** With the constants temporarily filled (`OFFICIAL_REGISTRY_USER = "dolos"`), the generated `dolos.toml` reads:

```toml
[mithril]
aggregator = "https://aggregator.release-mainnet.api.mithril.network/aggregator"
genesis_key = "5b3139312c…"
ancillary_key = "5b32332c…"

[stelae.registry]
user = "dolos"

[chain]
type = "cardano"
magic = 764824073
is_testnet = false
```

With them empty as they ship, the section is absent and nothing else changes. `init::tests::a_fresh_config_seeds_the_official_registry_and_no_password` holds both true — the generated config parses, round-trips to exactly `StelaeConfig::official()`, and never contains the word `password` — and `a_configured_password_overrides_the_compiled_in_one` covers the private-registry case and the fallback.

**3. Canonical harness against a basic-authed `registry:2`.**

```
running 9 tests
test a_layer_that_is_not_the_one_described_is_refused ... registry: registry:2 on 127.0.0.1:56820, plaintext, basic auth as "stelae"
...
test credentials_are_required ...
anonymous: registry error: Not authorized: url http://127.0.0.1:56845/v2/stelae/credentials/manifests/latest
the wrong password: registry error: Not authorized: url http://127.0.0.1:56845/v2/stelae/credentials/manifests/latest
ok
test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 7 filtered out; finished in 13.32s
```

The whole matrix, authenticated: `registry:2`, `registry:3` and `zot v2.1.20` — 9/9 for `stelae`, 5/5 for `dolos-snapshot`'s `publish`, 5/5 for `restore_registry`, on each, locally and in CI.

**4. Docs.** ADR-004's "What the transport requires of its host" now states the three shapes, the variable table and both refusals; its config sketch names `[stelae.registry]` and says where the password lives. `docs/content/configuration/schema.mdx` gains a `stelae` section documenting the user, the optional password and the environment variables.

## Verification

- `cargo clippy --workspace --all-targets --all-features` — clean
- `cargo test --workspace --all-targets` — pass
- `cargo test --workspace --all-features --exclude dolos-minibf --exclude dolos-minikupo --exclude dolos-trp` — pass
- the three `#[ignore]`d registry suites against all three matrix images — pass
- `cargo tree -p stelae -e normal --all-features` still matches no `dolos-*` package; no new dependency

## Notes for the reviewer

- **The read-only credential is a published secret by owner decision (2026-08-08).** Rotating it now means a release rather than an edit to every generated config; that risk is carried by the deployment plan, which owns rotation.
- **Half a pair is refused** as well as pair-plus-token. The plan named only the latter; a user with no password in the *environment* is the same class of operator mistake. (A user with no password in `dolos.toml` is the opposite — it is the seeded shape, and means "use the compiled-in one".)
- **`dolos snapshot publish` also passes the configured credentials.** They are read-only, so authenticating with them fails the push at the registry rather than a step earlier — the honest place for "these credentials cannot publish" to be said.
- **There is no `dolos bootstrap stelae` page under `docs/content/`** — a pre-existing gap, not opened by this change. The config section and the environment variables are documented in the schema page and ADR-004.

## Summary by CodeRabbit

* **New Features**
  * Added configurable registry authentication for anonymous access, bearer tokens, and HTTP Basic credentials.
  * Added registry credential settings, default-password behavior, and environment-variable overrides.
  * Publishing and restoring repositories now use configured authentication automatically.

* **Bug Fixes**
  * Added validation for conflicting or incomplete credentials.
  * Prevented authentication failures from appearing as empty repositories.

* **Documentation**
  * Documented registry authentication options, defaults, overrides, and configuration examples.

### DIFF
--- a/adrs/004_stelae_snapshots.md
+++ b/adrs/004_stelae_snapshots.md
@@ -240,14 +240,7 @@ The arithmetic is counted in layers, because layers are what the ceiling counts:
 #### What the transport requires of its host
 
 - **A process that opens a registry client must have installed a process-default rustls `CryptoProvider` first.** The transport ships no crypto backend of its own (`reqwest/rustls-no-provider`): the backend the client library would otherwise pick, `aws-lc-rs`, wants `cmake` on every build machine — the dependency this workspace already goes out of its way to avoid — so it stays out of the tree and the choice of provider moves to the program. In Dolos, `main()` installs `ring`. Omitting the install is a panic when the registry client opens, not a link error.
-- **Authentication is the host's decision, in one of three shapes.** The client is opened with credentials its caller supplies — anonymous, a bearer token, or an HTTP Basic pair — and never goes looking on its own: which identity a program authenticates as is that program's policy. What the protocol owns is the environment grammar, because it owns the variable names:
-
-  | Variable | Shape |
-  | --- | --- |
-  | `STELAE_REGISTRY_TOKEN` | bearer token |
-  | `STELAE_REGISTRY_USER` + `STELAE_REGISTRY_PASSWORD` | Basic pair |
-
-  An empty value is unset. A token and a pair set together is a **refusal**, not a precedence rule, and so is half a pair: an operator who exported both meant one of them, and a client that guessed would authenticate as an identity nobody chose — which on a registry whose credentials carry different capabilities is the difference between a publish and a 403 nobody can explain.
+- **Authentication is the host's decision, in one of three shapes.** The client is opened with credentials its caller supplies — anonymous, a bearer token, or an HTTP Basic pair — and never sources them itself. Which identity a program authenticates as is that program's credential policy, and where it keeps its credentials is that program's deployment: a protocol library that read an environment variable would be deciding both on its host's behalf, and naming the variable would freeze that decision into a published API. **So this specification names no environment variable and no configuration key**, and `stelae::oci::Options::auth` is the whole of the interface. Dolos's own answer is under "CLI and configuration" below.
 
   Anonymous remains legitimate and is what a genuinely public repository wants. It is not what a registry that authenticates every request wants, and that is the deployment Dolos is heading for: read access to a stele repository is free and identity-less, and still credentialed.
 
@@ -304,7 +297,16 @@ user = "…"                     # seeded by `dolos init`
 
 The official registry's read-only password is a **published secret**: it is what makes stele distribution free and identity-less while still authenticated. It is compiled into the binary rather than seeded into the file, so `dolos init` writes a `user` and no password and a rotation reaches every node that takes a release, instead of having to be found again in every generated `dolos.toml`. A node pointing at a private registry sets `password` and gets its own.
 
-A publisher's full-access pair is a real secret and belongs in the environment or a secret manager, never in this file — which is why the environment overrides both the file and the compiled-in default, so a node carrying the read-only user can publish without being edited first.
+A publisher's full-access pair is a real secret and belongs in the environment or a secret manager, never in this file, so `dolos` reads three variables of its own:
+
+| Variable | Shape |
+| --- | --- |
+| `STELAE_REGISTRY_TOKEN` | bearer token |
+| `STELAE_REGISTRY_USER` + `STELAE_REGISTRY_PASSWORD` | Basic pair |
+
+An empty value is unset. The environment overrides both the file and the compiled-in default, so a node carrying the read-only user can publish without being edited first. A token and a pair set together is a **refusal**, not a precedence rule, and so is half a pair: an operator who exported both meant one of them, and a client that guessed would authenticate as an identity nobody chose — which on a registry whose credentials carry different capabilities is the difference between a publish and a 403 nobody can explain.
+
+These are Dolos's variables and Dolos's rule, resolved in `dolos::common::stele_registry_auth`, which hands the answer to the transport as a value. Another host embedding `stelae` names its own, or none.
 
 ### Publisher pipeline
 
--- a/adrs/004_stelae_snapshots.md
+++ b/adrs/004_stelae_snapshots.md
@@ -240,7 +240,16 @@ The arithmetic is counted in layers, because layers are what the ceiling counts:
 #### What the transport requires of its host
 
 - **A process that opens a registry client must have installed a process-default rustls `CryptoProvider` first.** The transport ships no crypto backend of its own (`reqwest/rustls-no-provider`): the backend the client library would otherwise pick, `aws-lc-rs`, wants `cmake` on every build machine — the dependency this workspace already goes out of its way to avoid — so it stays out of the tree and the choice of provider moves to the program. In Dolos, `main()` installs `ring`. Omitting the install is a panic when the registry client opens, not a link error.
-- **Authentication is a bearer token from `STELAE_REGISTRY_TOKEN`, or anonymous.** Read once, when the registry client opens. Nothing else: registry credentials are a publisher's concern with a policy of its own.
+- **Authentication is the host's decision, in one of three shapes.** The client is opened with credentials its caller supplies — anonymous, a bearer token, or an HTTP Basic pair — and never goes looking on its own: which identity a program authenticates as is that program's policy. What the protocol owns is the environment grammar, because it owns the variable names:
+
+  | Variable | Shape |
+  | --- | --- |
+  | `STELAE_REGISTRY_TOKEN` | bearer token |
+  | `STELAE_REGISTRY_USER` + `STELAE_REGISTRY_PASSWORD` | Basic pair |
+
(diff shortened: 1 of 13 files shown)


scarmuega:
CI note: the first two attempts at `Test` were red on `daemon_syncs_for_preprod_full_implicit` (`tests/e2e/sync.rs:59`, `before=0, after=0`) — the 60-second live sync against `preprod-node.world.dev.cardano.org`. The `preview` scenarios passed in the same runs, `main` was green an hour earlier, and nothing here touches upstream sync, the peer path or `bootstrap relay`. It cleared on re-run without a code change. Recording it rather than passing over it: the scenario is network-dependent and there is no retry around it.

scarmuega:
Pushed `304c5cc1`, correcting two things in the shape `dolos init` writes.

**No commented-out template.** The generated `dolos.toml` had a commented `[stelae.registry]` block appended after serialization — a mechanism this repo uses nowhere else. Gone; the schema page documents the section.

**The password is compiled in, not seeded.** `dolos init` now writes the section with the official registry's `user` and nothing else, and `StelaeRegistryConfig::password()` falls back to `OFFICIAL_REGISTRY_PASSWORD` when the file names none. A password copied into every generated config has to be found again in every one of them; a compiled-in default rotates with a release. A private registry sets `password` and keeps its own, and the environment still overrides both.

Both constants still ship empty, so today init writes no section rather than one naming nobody. With them filled the generated file reads:

```toml
[stelae.registry]
user = "dolos"
```

scarmuega:
Pushed `2507ee1e`, which corrects a layering mistake I made rather than inherited.

`stelae` had no business naming an environment variable. A protocol library that reads one is choosing its host's credential policy, and publishing the name freezes that choice into its API — every program embedding the transport inherits three `STELAE_*` variables it never asked for and cannot rename. The justification in the code was circular: the crate owned the names because they started with `STELAE`.

- `TOKEN_ENV`, `USER_ENV`, `PASSWORD_ENV`, `Auth::from_env` and the two refusal variants leave `stelae` entirely. The crate no longer references `std::env` at all. What remains is the interface: `Options::auth` takes an `Auth` the caller constructs.
- `dolos_snapshot::registry::open` now takes an `Auth` rather than a config type, so the profile crate does not source credentials either.
- The names, the precedence and both refusals live in `dolos::common::stele_registry_auth` — the binary whose deployment they describe. It reads `&StelaeConfig` and the environment.
- ADR-004's protocol section states that credentials arrive from the caller and names no variable and no configuration key; the variable table moved to "CLI and configuration", where it is Dolos's answer rather than the format's.

Side effect worth having: the refusals are now plain miette errors raised in the binary, so the CLI prints the sentence once instead of three times through two error types.

```
Error:   × STELAE_REGISTRY_TOKEN and STELAE_REGISTRY_USER/STELAE_REGISTRY_PASSWORD
  │ are both set; registry credentials come from one of the two and which one
  │ was meant is not something to guess at — unset the one you did not mean
```

The resolution tests moved with the code, to `common::tests` in the binary. The registry suites keep proving that credentials reach the wire and decide the outcome, which is the part that belongs at that layer.

scarmuega:
Pushed `78611df1`. **Breaking:** `STELAE_REGISTRY_TOKEN`, `STELAE_REGISTRY_USER` and `STELAE_REGISTRY_PASSWORD` are gone, with no alias.

Checking how the rest of the binary handles environment variables settled this. clap's declarative `env` is used **nowhere** in the workspace — zero `#[arg(env = …)]`. The convention is the `config` crate's layer in `load_config`:

```rust
s = s.add_source(::config::Environment::with_prefix("DOLOS").separator("_"));
```

Every `RootConfig` field is already overridable as `DOLOS_<path>`, and `docs/content/configuration/introduction.mdx` teaches it. Before the OCI transport landed, the only `std::env::var` in non-test code across the whole workspace was… none — the `STELAE_REGISTRY_TOKEN` read introduced in #1170 was the single outlier, and this PR had been extending it.

So `DOLOS_STELAE_REGISTRY_USER` / `_PASSWORD` already worked, with no code. Verified before changing anything, against a live htpasswd registry with no `[stelae.registry]` in the file at all:

```
# DOLOS_STELAE_REGISTRY_PASSWORD correct
  ╰─▶ Registry error: … [OCI API error: manifest unknown]     ← past auth
# wrong
  ╰─▶ Not authorized: url http://…/v2/dolos/mainnet/manifests/latest
```

What changed:

- the three constants and the hand-rolled reader are gone; production code is back to zero hand-rolled environment reads;
- `[stelae.registry]` gains `token`, and `user`/`password` become optional, so the section can name one identity or the other and the bearer case is reachable as `DOLOS_STELAE_REGISTRY_TOKEN`;
- `stele_registry_auth` is a pure function of that section. The two refusals survive as validation over the resolved config — `token` with `user`, and `password` with no `user`:

```
Error:   × [stelae.registry] sets both `token` and `user`; a registry client
  │ authenticates as one identity and which one was meant is not something to
  │ guess at — drop the one you did not mean, or unset
  │ DOLOS_STELAE_REGISTRY_TOKEN / DOLOS_STELAE_REGISTRY_USER
```

- one test pins the mapping itself, building the same `config::Environment` source `load_config` does. A rename of either field, or a change to the prefix or separator, would otherwise fail silently at run time: the override would stop applying and a publisher would quietly authenticate as the read-only user.

ADR-004 now says Dolos introduces no environment variable of its own for this, and the schema page points at the existing env-override docs instead of listing bespoke names.

PR #1187, opened 2026-08-10
fix(stelae): a blob that already exists is a dedup, not a rename

scarmuega:
## What and why

`LayerSink::finish` publishes a layer by renaming its staging file onto
`blobs/sha256/<digest>` — where the digest is that of the very bytes being moved. If
that path already exists it already holds those bytes; content addressing says so. The
rename is a no-op in content terms and never the operation the situation calls for.

Doing it anyway costs two things:

- **I/O.** A blob that is already correct is rewritten. For a mainnet state shard that
  is hundreds of megabytes for no change.
- **A platform-divergent failure.** Narrower than #1167's body claimed, and the
  correction matters: `fs::rename` *does* replace an existing **file** on Windows (via
  `MoveFileExW` / `SetFileInformationByHandle`); the divergence std documents is about
  *directories*, which this code never renames. The real Windows divergence here is a
  sharing violation — every reader opens a blob with a plain `fs::File::open`, without
  `FILE_SHARE_DELETE`, so renaming over a blob that a concurrent `read_layer` or
  `stream_layer` holds open fails on Windows and succeeds on Unix.

Dedup removes the rename and both costs with it, without anyone having to reason about
share modes.

Nothing reaches this today, which is why it did not ride #1167: the sixteen state shards
each carry a distinct per-shard scope in their header record, so their blobs cannot
collide, and #1167's tests use separate steles or distinct scopes.

## What changed

**`crates/stelae/src/dir.rs`** — `LayerSink::finish` tests the destination before it
renames. If the blob is there, it stays; `Staging`'s `Drop` takes the duplicate staging
file, and the caller gets the same `WrittenLayer` it would have got had it been first
(`descriptor` and `digests` are computed from the record stream, never from the rename).
Otherwise the rename happens as before. Six lines of behaviour, and a doc-comment
section carrying the three things a reader needs:

- why an existing blob is a success rather than a rename (the Windows sharing violation
  included);
- why the existing blob is **trusted on its name and not re-read** — this writer computed
  that digest itself a microsecond earlier, and verifying would put a second full pass
  over a possibly multi-gigabyte file on the write path. That diverges from the
  neighbour, `blob_index`, which re-digests every blob it scans; the comment says why a
  fixture scanner with no manifest is not in the same position as a writer holding its
  own digest;
- why the check-then-rename window is an unguarded TOCTOU on purpose: the path *is* the
  digest, so the only racer that can create it is another writer of byte-identical
  content, and a lock would serialize every layer write to arbitrate between two callers
  who agree.

**`crates/stelae/tests/toy_profile.rs`** — one new test,
`a_blob_that_already_exists_is_deduplicated`. Same records, same header scope, same
stele, written twice. It asserts the second `WrittenLayer` equals the first's, that one
blob remains, that nothing is left staged beside `sha256/`, and that the layer still
reads back through both readers.

The evidence that **no bytes moved** is a doctored modification time: the first write's
blob is marked with a distinctive mtime, and since a rename replaces the file behind the
name — and the timestamp belongs to the file, not to the name — the mark surviving is
proof the second write left the blob alone. `File::set_times` is stable and
cross-platform, so this needs no `#[cfg]` seam. Reverting `dir.rs` alone makes the test
fail on exactly that assertion (`the blob was rewritten`).

## Out of scope, deliberately

- **No change to how readers open blobs.** Passing `FILE_SHARE_DELETE` is a different fix
  for a problem dedup dissolves, and would put a `#[cfg(windows)]` seam into a crate that
  has none.
- **No `crates/snapshot` change, no new public surface.** This is an internal branch in
  `finish`, not a new method.
- **No dedup across steles.** Blobs are addressed within one `SteleDir`; a registry-wide
  store is the OCI slice's problem.

## Verification

Run on `macos-14` (local, Rust 1.93):

| check | result |
| --- | --- |
| `cargo test --workspace --all-targets` | pass — every suite green, 0 failures |
| `cargo clippy --all-targets --all-features -- -D warnings` | pass |
| `cargo +nightly fmt --all -- --check` | pass |
| `cargo deny check advisories` | `advisories ok` |

`cargo deny` also emits three pre-existing `warning[yanked]` notices (`fjall`, `lsm-tree`,
`spin`) that are unrelated to this change and do not fail the check.

The goldens in `crates/stelae/tests/toy_profile.rs` are untouched and
`golden_digests_pin_the_encoding` still passes: this changes *when* bytes are written,
never what they are.

Done criterion 3 — the new test passing on all three CI platforms, which is where the
divergence this change exists for would show — is met by this PR's CI run: `Test
(ubuntu-latest)`, `Test (macos-14)` and `Test (windows-latest)` are all green, as are
`cargo clippy`, `cargo deny (advisories)`, `cargo fmt` and the three registry
round-trips.

## Note for the reviewer

`crates/stelae/tests/toy_profile.rs` is also being edited on `feat/stelae-publisher-commands`.
The new test here is appended between `both_write_paths_produce_the_same_layer` and
`sixteen_sinks_are_written_in_one_pass` and touches nothing else in the file, so whichever
lands second should be a trivial rebase.

## Trellis trail

- Plan: `plans/dolos-stelae-blob-dedup.md` (Brain/txpipe), `owner: org/coder`, now
  `status: active`.
- Done criterion: all five points met — 1, 2, 4 and 5 locally, 3 by this PR's CI.
- Escalations raised: none.
- Follow-up plan filed: none — no residue beyond what the plan covered.

### DIFF
--- a/crates/stelae/src/dir.rs
+++ b/crates/stelae/src/dir.rs
@@ -155,6 +155,37 @@ impl RecordSink for LayerSink {
     /// happens. A failure anywhere in here leaves the stele exactly as it was:
     /// the staging file is removed on the way out, the same as for a sink that
     /// was simply dropped.
+    ///
+    /// ## A blob that is already there is a success, not a rename
+    ///
+    /// The destination is named by the digest of the very bytes about to be
+    /// moved onto it, so a destination that already exists already holds those
+    /// bytes — that is what content addressing means. Renaming anyway rewrites
+    /// a file with its own contents: hundreds of megabytes of I/O for a mainnet
+    /// state shard, and on Windows a failure. Every reader here opens a blob
+    /// with a plain [`fs::File::open`], without `FILE_SHARE_DELETE`, so
+    /// replacing a blob that a concurrent [`SteleReader::stream_layer`] holds
+    /// open is a sharing violation there and silently fine on Unix. So the blob
+    /// on disk is kept, the staging file goes the way an abandoned one does,
+    /// and the caller gets exactly the [`WrittenLayer`] it would have got had
+    /// it been first: `descriptor` and `digests` are computed from the record
+    /// stream, never from the rename.
+    ///
+    /// The existing blob is **trusted on its name, not re-read**. This writer
+    /// computed that digest itself, out of the bytes it had just written, a
+    /// microsecond earlier, and re-reading a file that may be gigabytes to
+    /// confirm what content addressing already asserts would put a second full
+    /// pass on the write path. That is a deliberate divergence from the
+    /// neighbour: [`SteleDir::blob_index`] re-digests every blob it finds, at a
+    /// cost its own documentation calls out. It is a fixture scanner rebuilding
+    /// a map with no manifest to lean on and no knowledge of who wrote what,
+    /// which is not the position of a writer holding its own digest.
+    ///
+    /// The window between the test and the rename is a TOCTOU, and an unguarded
+    /// one on purpose: the path *is* the digest, so the only racer that can
+    /// create it is another writer of byte-identical content, and both orders
+    /// leave the same blob. A lock here would serialize every layer write to
+    /// arbitrate between two callers who agree.
     fn finish(self) -> Result<WrittenLayer, Error> {
         let Self {
             sequence,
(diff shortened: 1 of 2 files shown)


scarmuega:
On CodeRabbit's one actionable comment — replace check-then-rename with an atomic no-replace publish: not taken, deliberately, and the doc comment already states why.

There is no portable no-replace rename in std. Getting one means `RENAME_NOREPLACE`/`linkat` on Linux, `renamex_np` on macOS and `MoveFileExW`/`CreateHardLinkW` on Windows — a `#[cfg]` seam in a crate that has none, which this change's scope explicitly refuses (the whole point of dedup here is to dissolve the Windows sharing-violation case *without* anyone having to reason about share modes).

And the race it would close is not a race that can hurt: the destination path **is** the digest of the bytes being published, so the only writer that can win it is one writing byte-identical content. Both orders leave the same blob under the same name, and the loser's `WrittenLayer` is identical either way. That is why the window is left unguarded rather than locked or made atomic, and the `finish` doc comment says so in as many words.

Worth noting the residual case the current code handles the same way it did before this change: if the blob appears *after* the existence test, the rename replaces a byte-identical file — exactly the pre-change behaviour, so nothing regresses. The narrower window is a strict improvement over the wide one that existed before.

PR #1188, opened 2026-08-10
feat(stelae): give a publisher digest, verify and inspect

scarmuega:
Phase 4, publisher productization: the three read/verify commands and incremental detection. Implements the Trellis plan `plans/dolos-stelae-publisher-commands.md` ("Stelae — a publisher can ask what it has, and a verifier can reproduce it") of the Brain/txpipe domain, contexts `solution/stelae` + `solution/dolos`.

## What changed

**`crates/stelae`** — a `Discarding` `SteleWriter`: every layer framed, hashed and compressed exactly as a publish would — zstd runs in full, because the bug class this exists to catch only appears when the same bytes go through the same pipeline twice — and the bytes go to `std::io::sink`. What comes back is the identity.

**`crates/snapshot`** —
- `history_for` and `same_network` moved from `registry.rs` (feature-gated) into `export.rs`: the contiguity rule is one function shared between the registry publisher and a verifier that has no registry.
- `export::Following` reads a predecessor's canonical inscription off disk and extends it by the publisher's own rule (`digest --chain-from`); a re-encoded copy is refused, since a stele is chained to by the digest of its own bytes.
- `export::Attested` carries a published history verbatim; `export::verify_reproduction` rebuilds every layer through the discarding writer and compares descriptors kind by kind and scope by scope before comparing the inscription digest, with a cheap sequence refusal before the walk.
- `export::Standing` reads where a node stands against a repository (empty / up to date / next / ahead-with-distance) out of the two numbers a publish already has.
- `registry::verify` streams every blob end to end — blob digest and compressed size against the manifest, `diffId`, uncompressed size and record count against the inscription — and names the offending layer's kind and scope on failure (`Error::LayerVerification`). `registry::inspect` reads the manifest and the config blob and no layer.

**`src/bin/dolos/snapshot/`** — three new modules:
- `digest` — canonical inscription + sha256 from local stores and no registry; `--chain-from FILE`, `--output FILE`; stdout is the byte-exact document (no trailing newline), the report goes to stderr.
- `verify` — transport checks by default; `--reproduce` opt-in, and the help says it costs what a publish costs. Output states plainly: digests only; signatures and chain provenance are phase 5.
- `inspect` — sequence, position, profile, compression, history depth with first/last entries, one line per layer with the compressed size the manifest carries, totals; `--json` emits the canonical inscription verbatim. No signers column until there are signatures to list; the help says which.
- `publish` — a repository already at the node's sequence reports "nothing to publish" and exits zero; `--require-new` makes that case an error; a node further ahead is still refused, now with the distance alongside both sequences. `EpochRange` and the plan report are hoisted to `snapshot/mod.rs` so `publish`, `digest` and `verify` share one epoch parser and one report.

**Scope decisions honored**: no `sign` / signature verification, no CI workflow, no gap policy (the refusal stands, its message now names the distance), no `[snapshot] source`, no tag pruning, no progress reporting, no `--scratch-dir`.

## Done criteria

1. ✅ `digest` and `publish --output-dir` produce byte-identical canonical inscriptions — `a_discarding_export_reproduces_what_a_publish_stores` (records-bearing harness store, compared on the canonical bytes) and `a_discarding_writer_reproduces_what_a_directory_stores` (toy profile: every descriptor field, blob digest, compressed size, seal).
2. ✅ `digest --chain-from` reproduces a stele published **with reuse on** — `a_stele_published_with_reuse_is_reproduced_from_the_stores` (3 layers inherited, digest reproduced from stores alone; chained onto nothing is a different digest).
3. ✅ `verify --repo` passes against a fresh stele, and refuses with the offence named: a blob that is not the layer (streamed check, layer kind + scope named), a manifest `diffId` annotation disagreeing (refused at the pull, layer position + both identities named), a history that skips a sequence (refused at parse, gap named). Tampered artifacts are planted through the raw distribution API — the transport refuses to write any of them.
4. ✅ `verify --reproduce` passes against the store the stele was published from and fails against a store standing at a different epoch — the failure costs a comparison of two sequences, not hours.
5. ✅ `inspect` lists every layer with the manifest's compressed size (inherited layers included), sizes sum to the manifest total, and its canonical JSON round-trips through `Following::read` — the exact `digest --chain-from` path — chaining to the successor's published digest.
6. ✅ `publish --repo` at the node's sequence: nothing to publish, exit zero; `--require-new` non-zero; three ahead refused with both sequences and "3 sequences ahead" in the message — `a_repository_is_read_as_empty_current_next_or_ahead`, `a_gap_names_the_distance_alongside_both_sequences`, `a_publisher_can_ask_where_it_stands`.
7. ✅ Registry end-to-end tests `#[ignore]`d and **run** — output quoted below.
8. ✅ Full gate — results below.

## Verification

- `cargo test` — full workspace: 15 suites, 0 failures.
- `cargo +nightly fmt --all -- --check` — clean.
- `cargo deny check advisories` — `advisories ok`.
- `cargo tree -p stelae -e normal --all-features` — matches nothing `^dolos(-|$)`.

### Ignored registry suites, run against a spawned `registry:2`

`cargo test -p dolos-snapshot --features oci --test snapshot_verify -- --ignored --nocapture`:

```
diffId disagreement: stelae error: manifest disagrees with the inscription: layer 0 is annotated sha256:94455e3ed9f716bea425ef99b51fae47128769a1a0cd04244221e4e14631ab83 and the inscription describes sha256:0329ec9ba1660678676c18209478eed13ef607e234d46e5de275674563dd8ede there
corrupted blob: the blocks layer at {"endSlot":86399,"epoch":0,"startSlot":0} failed verification: stelae error: malformed layer header record: unexpected type string at position 0: expected array
skipped sequence: stelae error: history invariant violated: gap between sequence 0 and sequence 2
reproduced sha256:ef955801a19cee27e7a6069f7b653cda2b2402fc167746b2e8059f8a41b1881f == published sha256:ef955801a19cee27e7a6069f7b653cda2b2402fc167746b2e8059f8a41b1881f; a store at another epoch: the reproduction does not match the published stele — sequence: the published stele is sequence 2 and these stores stand at sequence 1; a reproduction runs over stores at the epoch the stele was published from
inspected 22 layers, 7830 compressed bytes; epoch-1's document chained to sha256:ef955801a19cee27e7a6069f7b653cda2b2402fc167746b2e8059f8a41b1881f
verified latest = sha256:ef955801a19cee27e7a6069f7b653cda2b2402fc167746b2e8059f8a41b1881f (22 layers, 7830 compressed bytes) and epoch-1 = sha256:f9523cbfed6dc7fad48f0d1452948dfe8988218b63bb8d1e2c60cfadc859321a

test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.68s
```

`cargo test -p dolos-snapshot --features oci --test publish -- --ignored --nocapture`:

```
publish 1: built 19, reused 0, uploaded 19 (7666 bytes)
publish 2: built 19, reused 3 (3390 bytes not moved), uploaded 19 (4440 bytes)
published (3 layers inherited) sha256:ef955801a19cee27e7a6069f7b653cda2b2402fc167746b2e8059f8a41b1881f == reproduced from stores sha256:ef955801a19cee27e7a6069f7b653cda2b2402fc167746b2e8059f8a41b1881f
inherited sha256:ef955801a19cee27e7a6069f7b653cda2b2402fc167746b2e8059f8a41b1881f == rebuilt sha256:ef955801a19cee27e7a6069f7b653cda2b2402fc167746b2e8059f8a41b1881f

test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3.33s
```

`cargo test -p dolos-snapshot --features oci --test restore_registry -- --ignored --nocapture`:

```
test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.27s
```

`cargo test -p stelae --features oci --test oci -- --ignored`: `10 passed; 0 failed` (13.94s).

## Notes for review

- **`digest`/`inspect --json` stdout carries no trailing newline.** The document is bytes: `digest > stele.json` has to hash to the reported identity, and `--chain-from` refuses anything that is not the canonical encoding itself. The plan report moved to stderr for the same reason; `tests/snapshot_publish.rs` was updated to read it there.
- **`verify --reproduce` has not been run against a production-published stele** — it needs a node whose stores stand at the published epoch, which this environment does not have. Per the plan's risk section, a failing `--reproduce` against our own published stele is a finding for org/founder, not a comparison to adjust; the determinism job of `dolos-stelae-publisher-pipeline` is where that run lives.
- ADR-004's stale "checked in CI" claims about the `cargo tree` boundary were trimmed to match reality (the CI guard plan is back in draft); the check ran by hand as part of the gate here.

## Trellis trail

- Plan: `plans/dolos-stelae-publisher-commands.md` (Brain/txpipe) — done criterion **met**; plan stays `active`, retirement is the owner's verdict.
- Escalations: none.
- Follow-up: `plans/dolos-stelae-sign.md` (draft) — Phase 5 signing, the one deferral in this plan with no plan to point at; this PR now promises "phase 5" in `verify`'s output, `inspect`'s help and the module docs.

## Summary by CodeRabbit

- **New Features**
  - Added `snapshot digest` to reproduce snapshot identities without writing repository data.
  - Added `snapshot inspect` for metadata and canonical JSON output without downloading layers.
  - Added `snapshot verify`, including optional local reproduction checks.
  - Added predecessor chaining, attested histories, and epoch-range filtering.
  - Added repository status reporting and `--require-new` publishing protection.

- **Bug Fixes**
  - Improved detection and reporting of broken histories, mismatched layers, networks, and corrupted snapshot data.

- **Documentation**
  - Updated snapshot architecture documentation.

### DIFF
--- a/adrs/004_stelae_snapshots.md
+++ b/adrs/004_stelae_snapshots.md
@@ -106,7 +106,7 @@ None of those four goals is Cardano-specific, and neither are the mechanisms tha
 
 10. **Specify the mechanism as a Dolos feature, without a protocol name or profile boundary** (the shape of this ADR before the Stelae amendment)
     - Pros: one crate, one vocabulary, no extension machinery to design or test.
-    - Cons: a third-party publisher has no collision-free namespace and would have to fork the spec; the Dolos context absorbs decisions (framing, attestation, transport) unrelated to a data node; and extraction later means renaming every media type, tag and identifier already published. The boundary costs one crate and one CI check today and is irreversible-cheap only before implementation starts.
+    - Cons: a third-party publisher has no collision-free namespace and would have to fork the spec; the Dolos context absorbs decisions (framing, attestation, transport) unrelated to a data node; and extraction later means renaming every media type, tag and identifier already published. The boundary costs one crate today and is irreversible-cheap only before implementation starts.
 
 ## Implementation Details
 
@@ -246,7 +246,7 @@ The arithmetic is counted in layers, because layers are what the ceiling counts:
 
 ### Code layout
 
-Two crates, both workspace members. The split is the protocol/profile boundary made mechanical: **`cargo tree -p stelae` must contain no `dolos-*` package**, checked in CI, so extracting the protocol later is a directory move rather than a refactor.
+Two crates, both workspace members. The split is the protocol/profile boundary made mechanical: **`cargo tree -p stelae` must contain no `dolos-*` package**, so extracting the protocol later is a directory move rather than a refactor.
 
 ```
 crates/stelae/            # package `stelae` — protocol, zero dolos deps
@@ -335,7 +335,7 @@ Steps 1–2 and the fetch/verify half of steps 4–5 are protocol code; the stor
 
 ### Development phases
 
-**1a. Stelae core** — `crates/stelae`: framing, inscription (schema, JCS, digest, history invariant), the `Profile` trait and naming rules, streaming digest/compression, signatures. Verified by CBOR-seq roundtrip and write→read→write byte-identity property tests, a JCS inscription golden test, history-invariant tests (gap/duplicate/out-of-order → reject), fail-closed tests (unknown generic key, unknown profile, higher profile major), a toy non-Dolos profile exercising the full path, and the `cargo tree -p stelae` boundary check.
+**1a. Stelae core** — `crates/stelae`: framing, inscription (schema, JCS, digest, history invariant), the `Profile` trait and naming rules, streaming digest/compression, signatures. Verified by CBOR-seq roundtrip and write→read→write byte-identity property tests, a JCS inscription golden test, history-invariant tests (gap/duplicate/out-of-order → reject), fail-closed tests (unknown generic key, unknown profile, higher profile major), and a toy non-Dolos profile exercising the full path.
 
 **1b. Dolos profile core** — `crates/snapshot`: `DolosProfile`, layer readers/writers, and the three trait additions with backend impls and adapter enums. Verified by per-layer roundtrip unit tests and golden-digest tests (fixed input → asserted sha256, catching encoding drift).
 
--- a/crates/snapshot/src/export.rs
+++ b/crates/snapshot/src/export.rs
@@ -647,7 +647,13 @@ where
 ///
 /// The sequence is checked before a single record is walked. A plan standing
 /// at a different epoch than the published stele cannot match it, and finding
-/// that out should not cost the hours of compression a full walk does.
+/// that out should not cost the hours of compression a full walk does. The
+/// network is checked on the same terms, and it is not implied by the sequence:
+/// epoch numbers collide across chains, so a preprod stele and a mainnet store
+/// can stand at the same sequence and still have nothing to say to each other.
+/// [`Attested`] takes the published history verbatim and so — unlike
+/// [`Following::new`] — checks nothing about it; this is where that check
+/// lands.
 pub fn verify_reproduction<A, S, I>(
     published: &Inscription,
     plan: &Plan,
@@ -671,6 +677,11 @@ where
             ),
         });
     }
+
+    // Same reasoning as the sequence check, and not covered by it: a store on
+    // another chain cannot reproduce this stele, and epoch numbers collide
+    // across networks, so same-sequence-different-chain is reachable.
+    same_network(published, plan)?;
 
     let reproduced = export(
         &stelae::Discarding,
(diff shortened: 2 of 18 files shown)


scarmuega:
Review pass applied in 5c127046.

**Applied (5)**

- Explicit stdout flush in `snapshot digest` and `snapshot inspect --json`. The canonical inscription has no trailing newline, so a line-buffered stdout held its tail and the implicit flush at exit discarded the error — `dolos snapshot digest > stele.json` on a full disk or a closed pipe exited zero with a truncated document. The `--output` path already propagated via `fs::write`.
- `verify_reproduction` now calls `same_network` right after the sequence check, before the walk. Epoch numbers collide across chains, and `Attested::of` (unlike `Following::new`) checks nothing, so same-sequence/different-chain was reachable and cost a full reproduction to discover.
- Unit test for `Following::read`'s canonical-form refusal — reordered keys and a trailing newline, both parseable, both `MalformedInscription`. Every call site in the tree passes canonical bytes, so the branch had no coverage.
- `snapshot/mod.rs` heading no longer names a command count; `restrict` has three callers.
- `Distribution::request` in `snapshot_verify` bounds connect/read/write, matching `wait_until_ready` in the registry fixture. A wedged registry was a CI hang rather than a failure.

**Declined (2)**

- `registry.rs:443` unreadable compressed size — unreachable at that line: `stream_layer` resolves the layer through the same `layer_of` (returning `LayerNotFound` if absent) and pulls the blob under a ceiling clamped from `descriptor.size`, so an absent or negative size fails before the check would run.
- `Digest::ALGORITHM` in `tests/export.rs` — a test pinning on-disk layout should trip on an algorithm change rather than follow it; the `"sha256"` literal is also the local majority in that file (4 sites vs 2).

The shared `layer_sink` prologue across `dir.rs`/`oci.rs`/`transport.rs` is deferred: two of the three copies predate this PR, so extracting it is out of this branch's scope.

Gate re-run green: `cargo test` (workspace, 0 failures), `cargo clippy --all-targets --all-features -- -D warnings`, `cargo +nightly fmt --all -- --check`, `cargo deny check advisories` (ok; the three pre-existing yanked-crate warnings unchanged), `cargo tree -p stelae -e normal --all-features` still matches nothing `^dolos(-|$)`. The `#[ignore]`d registry suites were re-run against a spawned `registry:2`: `snapshot_verify` 6/6, `publish` 7/7, `restore_registry` 5/5, with the same published identity as before (`sha256:ef955801…1b1881f`).

PR #1191, opened 2026-08-11
feat(stelae): stage where the operator said, in both directions

scarmuega:
Plan: `plans/dolos-stelae-publisher-operability-scratch-dir.md` (Trellis root `Brain/txpipe`, owner
`org/coder`), the first piece of the `dolos-stelae-publisher-operability` umbrella. Base: `main` at
`0ff6e17d` (#1188), which rewrote `crates/snapshot/src/registry.rs`.

## The defect

`registry::open` passed `scratch_dir: None` to `stelae::oci::Options`. `Shared::scratch` is reached
from `layer_sink` on the way up and `pull_blob_file` on the way down, so every layer a publish built
and every layer a restore pulled was staged in the platform temporary directory — not always on a
volume with room for sixteen mainnet state shards. `stelae::oci` honoured the option all along in
both directions; nothing set it.

## What changed

- `crates/snapshot/src/registry.rs` — `open` takes the staging directory as a required `PathBuf`.
  Not an `Option`: a parameter nobody can omit cannot grow another `None`.
- `src/bin/dolos/common.rs` — `stele_scratch_dir`, one default (`<storage.path>/scratch`) for all
  four commands that open a repository, with a unit test on the default and the override.
- `--scratch-dir` on `snapshot publish` (conflicting with `--output-dir`, which stages nothing) and
  on `bootstrap stelae`. `snapshot verify` and `snapshot inspect` take the default with no flag.
- `crates/core/src/config.rs`, `docs/content/configuration/schema.mdx`, `AGENTS.md` — `storage.path`
  documents `scratch/`.

The default is inside `storage.path` rather than beside it: on a dedicated data mount a sibling
lands on the parent filesystem, which is the small root volume the default exists to keep bytes off.
`tempfile_in` unlinks at creation, so `scratch/` reads as empty even mid-transfer. Decided by
`org/founder`; the documentation change is part of the deliverable because it changes what an
operator sizing that volume has to count.

## Tests

The unit test on `stele_scratch_dir` covers the default and that an explicit `--scratch-dir` is
taken literally, relative paths included. `staging_stays_in_the_scratch_directory_and_leaves_nothing`
in `crates/stelae/tests/oci.rs` already covers the transport honouring the option in both
directions. No CLI-level suite: a container-spawning integration test is out of proportion to a
`PathBuf` argument on two commands.

The shared registry fixture now names a `TempDir` of its own, so the `#[ignore]`d registry suites
stop staging in `/tmp` too.

## Verification

On the committed toolchain (`cargo +1.93`), macOS aarch64.

```
$ cargo test --workspace --all-targets                      # 0 failed
$ cargo clippy --all-targets --all-features -- -D warnings   # clean
$ cargo +nightly fmt --all -- --check                        # clean
$ cargo deny check advisories                                # ok
$ cargo tree -p stelae -e normal --all-features | grep -E 'dolos(-|$| v)'   # no match
```

Help output and the flag conflict, checked by hand:

```
$ dolos snapshot publish --output-dir /tmp/x --scratch-dir /tmp/y
error: the argument '--output-dir <OUTPUT_DIR>' cannot be used with '--scratch-dir <DIR>'
```

The `#[ignore]`d registry suites were run against `registry:2` in Docker, since this PR changes the
shared fixture they all use:

```
$ cargo test -p stelae --all-features --test oci -- --ignored                 # 10 passed
$ cargo test -p dolos-snapshot --features oci --test publish -- --ignored     # 7 passed
$ cargo test -p dolos-snapshot --features oci --test restore_registry -- --ignored  # 5 passed
$ cargo test -p dolos-snapshot --features oci --test snapshot_verify -- --ignored   # 6 passed
```

## Notes

- Not in this PR, per the umbrella's sequencing: preflight sizing, resumption, progress reporting.
- Follow-up filed as `plans/dolos-stelae-scratch-dir-diagnostics.md`: an unusable `--scratch-dir`
  fails with a bare `Not a directory (os error 20)` because neither `create_dir_all` nor
  `tempfile_in` carries the path into its error. That is `stelae::oci`'s to fix.
- Reported rather than fixed: `crates/snapshot/tests/snapshot_verify.rs`, added by #1188, is
  `#[ignore]`d with nothing running it — neither in `registry.yml` nor in AGENTS.md. Run by hand
  here and green.

## Summary by CodeRabbit

* **New Features**
  * Added configurable scratch storage for temporary snapshot and registry transfer data.
  * Added `--scratch-dir` options for snapshot publishing and stelae bootstrapping.
  * Scratch storage defaults to the configured storage volume’s `scratch/` directory.
  * Snapshot inspection, verification, publishing, and restoration now use consistent staging storage.
  * Publishing prevents conflicting use of `--scratch-dir` and `--output-dir`.

* **Documentation**
  * Documented the scratch directory layout, supported workflows, overrides, and potentially large temporary storage requirements.


scarmuega:
Review pass (`86fe93d7`): CodeRabbit's one comment applied, and it was right — `snapshot verify` streams every blob through a staged file and, unlike the publish and the restore, has no `--scratch-dir` to point elsewhere, so an operator sizing that volume off these documents would have sized it for two of the three commands that use it. Fixed in all three places it names: `docs/content/configuration/schema.mdx`, `crates/core/src/config.rs`'s `StorageConfig::path`, and AGENTS.md's storage layout.

No code change; the gate was re-run green (`cargo test --workspace --all-targets`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo +nightly fmt --all -- --check`).

crates/core/src/config.rs
```diff
@@ -546,6 +546,12 @@ pub struct StorageConfig {
     pub version: StorageVersion,
 
     /// Root directory for storage files.
+    ///
+    /// Besides the stores it holds `scratch/`, where a stele transfer over an
+    /// OCI registry stages the layers it is moving. That is the default
+    /// because this volume is already sized for the data; a mainnet transfer
+    /// stages gigabytes, which the platform temporary directory cannot be
+    /// assumed to fit.
```
scarmuega: this is excessive docs about scratch, not the main focus of this field.

PR #1195, opened 2026-08-12
feat(minibf): add `/accounts/{stake_address}/addresses/total`

### LINKED ISSUE #1066: minibf: add `/accounts/{stake_address}/addresses/total`
Implement the Blockfrost `/accounts/<account>/addresses/total` endpoint in the `minibf` crate.

### Effort

M — Medium (data exists, needs derivation)

> ⚠️ **TODO**: @vladimirvolek says this assessment is not entirely accurate. Dolos has `max_history = XXX`. We need to update the data model, indexer and change is also outside of the `minibf` crate.

### Approach

Sent/received aggregation over `stake` tx index \+ per-input resolution. Highest-value gap.

### Why now

2.9501% of observed traffic.

vladimirvolek:
resolves: https://github.com/txpipe/dolos/issues/1066

## Summary by CodeRabbit

* **New Features**
  * Added the `/accounts/{stake_address}/addresses/total` endpoint.
  * Returns lifetime received and sent asset totals, along with the transaction count, for a stake address.

* **Documentation**
  * Documented the new endpoint and clarified that totals reflect retained archive history when `sync.max_history` is configured.
  * Updated Mini Blockfrost endpoint coverage information.

### DIFF
--- a/crates/minibf/src/routes/accounts.rs
+++ b/crates/minibf/src/routes/accounts.rs
@@ -8,6 +8,8 @@ use axum::{
 use blockfrost_openapi::models::{
     account_addresses_assets_inner::AccountAddressesAssetsInner,
     account_addresses_content_inner::AccountAddressesContentInner,
+    account_addresses_total::AccountAddressesTotal,
+    account_addresses_total_received_sum_inner::AccountAddressesTotalReceivedSumInner,
     account_content::AccountContent,
     account_delegation_content_inner::AccountDelegationContentInner,
     account_history_content_inner::AccountHistoryContentInner,
@@ -42,7 +44,7 @@ use pallas::ledger::primitives::conway::Certificate as ConwayCert;
 use crate::{
     error::Error,
     inputs::{for_each_touched_output, InputDeps, InputResolver},
-    mapping::{self, bech32_drep, bech32_pool, IntoModel},
+    mapping::{self, bech32_drep, bech32_pool, AssetTotals, IntoModel},
     pagination::{Order, Pagination, PaginationParameters},
     Facade,
 };
@@ -297,6 +299,150 @@ where
     }
 
     Ok(Json(items))
+}
+
+/// Fold one block's txs into an account's lifetime totals.
+///
+/// Produced outputs are received and resolved inputs are sent, the same split
+/// `/addresses/{address}/total` makes for a single address; a tx counts once
+/// however many of the account's addresses it touches.
+async fn sum_account_block_txs<D>(
+    domain: &Facade<D>,
+    deps: &mut InputDeps,
+    account: &[u8],
+    block: &[u8],
+    received: &mut AssetTotals,
+    sent: &mut AssetTotals,
+    tx_count: &mut usize,
+) -> Result<(), StatusCode>
+where
+    D: Domain + Clone + Send + Sync + 'static,
+{
+    let block = MultiEraBlock::decode(block).map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+
+    let txs = block.txs();
+
+    let mut resolver = deps.prepare(domain, txs.iter()).await?;
+
+    for tx in txs.iter() {
+        let mut matched = false;
+
+        for (_, output) in tx.produces() {
+            let address = output
+                .address()
+                .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+
+            if address_belongs_to_account(&address, account) {
+                received.add_output(&output);
+                matched = true;
+            }
+        }
+
+        for input in tx.consumes() {
+            if let Some(output) = resolver.resolve(&input)? {
+                let address = output
+                    .address()
+                    .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+
+                if address_belongs_to_account(&address, account) {
+                    sent.add_output(&output);
+                    matched = true;
+                }
+            }
+        }
+
+        if matched {
+            *tx_count += 1;
+        }
+    }
+
+    Ok(())
+}
+
+fn account_amounts(totals: AssetTotals) -> Vec<AccountAddressesTotalReceivedSumInner> {
+    totals
+        .into_amounts()
+        .into_iter()
+        .map(|amount| AccountAddressesTotalReceivedSumInner {
+            unit: amount.unit,
+            quantity: amount.quantity,
+        })
+        .collect()
+}
+
+/// `GET /accounts/{stake_address}/addresses/total`: lifetime sums and tx count
+/// across every address of an account.
+///
+/// The totals are folded from the archive on each request rather than kept as
+/// state: a per-account asset breakdown is unbounded — one row grows with every
+/// distinct asset the account ever touched — and it would ride along in every
+/// stele and every state rebuild. The account therefore pays the same full
+/// scan `/addresses/{address}/total` already pays for a single address, which
+/// also means the answer only covers the history the archive still holds.
+pub async fn by_stake_addresses_total<D>(
+    Path(stake_address): Path<String>,
+    State(domain): State<Facade<D>>,
+) -> Result<Json<AccountAddressesTotal>, Error>
+where
+    Option<AccountState>: From<D::Entity>,
+    D: Domain + Clone + Send + Sync + 'static,
+{
+    let network = domain.get_network_id()?;
+    let account_key = parse_account_key_param(&stake_address, network)?;
+
+    if !domain.cardano_entity_exists::<AccountState>(account_key.entity_key.as_slice())? {
+        return Err(StatusCode::NOT_FOUND.into());
+    }
+
+    let account = account_key.address.to_vec();
+    let end_slot = domain.get_tip_slot()?;
+
+    let stream = domain
+        .query()
+        .blocks_by_stake_stream(&account, 0, end_slot, SlotOrder::Asc);
+
+    let mut received = AssetTotals::default();
+    let mut sent = AssetTotals::default();
+    let mut tx_count: usize = 0;
+    let mut deps = InputDeps::default();
+
+    let mut stream = Box::pin(stream);
+
+    while let Some(res) = stream.next().await {
+        let (_slot, block) = res.map_err(|err| {
+            tracing::error!(?err);
+            StatusCode::INTERNAL_SERVER_ERROR
+        })?;
+
+        let Some(block) = block else {
+            continue;
+        };
+
+        sum_account_block_txs(
+            &domain,
+            &mut deps,
+            &account,
+            &block,
+            &mut received,
+            &mut sent,
+            &mut tx_count,
+        )
+        .await?;
+    }
+
+    let stake_address = account_key
+        .address
+        .to_bech32()
+        .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+
+    let model = AccountAddressesTotal {
+        stake_address,
+        received_sum: account_amounts(received),
+        sent_sum: account_amounts(sent),
+        tx_count: tx_count as i32,
+    };
+
+    Ok(Json(model))
 }
 
 pub async fn by_stake_utxos<D>(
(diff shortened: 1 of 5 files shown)


michalrus (commented):
@vladimirvolek there are some conflicts in `crates/cardano/src/model/mod.rs`

crates/cardano/src/roll/accounts.rs
```diff
@@ -61,6 +119,18 @@ impl BlockVisitor for AccountVisitor {
         Ok(())
     }
 
+    fn visit_tx(
```
slowbackspace: 🟡 `visit_tx` here flushes the **previous** tx's buffer, not the current one: the hook runs before this tx's `visit_input`/`visit_output`, so `tx_activity` still holds tx N-1, and the block-end `flush()` covers the last tx. That ordering is not part of the `BlockVisitor` contract — every other implementor treats `visit_tx` as "this tx begins" — so any second driver (or a well-meaning refactor of `crawl`) that doesn't interleave the calls this way silently merges txs into one `AccountActivityRecord` and corrupts `tx_count`.

Suggest either computing the record entirely inside `visit_tx` (it already receives the tx and the resolved utxos map, so no cross-call state is needed), or documenting the ordering on the trait and renaming this to `flush_previous_tx_activity`.

crates/cardano/src/model/activity.rs
```diff
@@ -0,0 +1,341 @@
...
+
+    pub fn sent_lovelace(&self) -> u128 {
+        u128::from_be_bytes(self.sent_lovelace_bytes)
+    }
+}
+
+/// Cumulative native-asset activity for all addresses associated with an
+/// account, keyed by the same stake credential as `AccountActivity`.
+#[derive(Debug, Encode, Decode, Clone, Default, PartialEq, Eq)]
+pub struct AccountAssetActivity {
```
slowbackspace: 🟡 This map holds every asset unit the account ever touched, and the whole entity is decoded + re-encoded on each block that moves any of the account's assets — O(assets-ever-seen) per block, both at tip and during replay. For token/NFT-spammed hot wallets (tens of thousands of distinct units) that is a multi-MB codec pass per touching block.

Fine to ship as-is in my view, but worth a follow-up issue. Note for that issue: the naive fix (one entity per `cred ++ unit`) does **not** work — `EntityKey` is fixed 32 bytes and `From<&[u8]>` silently truncates, and the CBOR credential alone fills all 32. A workable shape is sharding: key = `cred_hash(28) ++ variant(1) ++ shard_byte(1) ++ pad(2)` with `shard = hash(unit) & 0xFF`; the endpoint reads the account's shard range via `iter_entities`, and per-block work drops to the shards actually touched.

crates/cardano/src/model/activity.rs
```diff
@@ -0,0 +1,341 @@
...
+    utils::Bytes,
+};
+use pallas::ledger::primitives::StakeCredential;
+use serde::{Deserialize, Serialize};
+
+use super::FixedNamespace as _;
+
+/// Cumulative lovelace activity for all addresses associated with an account.
+#[derive(Debug, Encode, Decode, Clone, Default, PartialEq, Eq)]
+pub struct AccountActivity {
```
slowbackspace: 🟡 Two core primitives already exist for this:

1. `dolos_core::cbor::U128` is the shared u128-as-16-bytes CBOR encoding — its doc says it exists "to give a single definition to an encoding that would otherwise be repeated as a private module in each crate", and `EpochState.output` already uses it. Declaring these fields as `cbor::U128` deletes `add_to_quantity`, both accessors, and the manual `from_be_bytes` conversions in minibf. The namespace is new, so there is no stored-data compat concern.
2. `cred_key` below re-implements `credential_to_key` (exported from `model/pending.rs`; rupd carries another copy). `NsKey::from((ns, credential_to_key(cred)))` keeps the key encoding defined once — three copies drifting means account lookups silently miss.

crates/cardano/src/roll/accounts.rs
```diff
@@ -75,6 +145,14 @@ impl BlockVisitor for AccountVisitor {
             return Ok(());
         };
 
+        // pointer addresses are excluded to match the account endpoints
+        if !is_pointer {
```
slowbackspace: 🟢 Two notes on the pointer exclusion:

- The policy is decided inline at two call sites (here and in `visit_output`) with the justifying comment only here. The established pattern in this same visitor pushes `is_pointer` into the delta (`ControlledAmountInc`/`Dec` decide in `apply`), keeping the rule at one point so the two guards cannot drift.
- It is also a real (if tiny) Blockfrost divergence worth documenting: db-sync resolves pointer addresses to `stake_address_id`, so BF includes their funds in totals, while dolos returns zeros for pointer-only credentials. Deliberate and consistent with the other dolos account endpoints — document-only.

crates/cardano/src/model/activity.rs
```diff
@@ -0,0 +1,341 @@
...
+        for (unit, amount) in &self.received_assets {
+            add_to_assets(&mut entity.received_assets, unit, *amount);
+        }
+
+        for (unit, amount) in &self.sent_assets {
+            add_to_assets(&mut entity.sent_assets, unit, *amount);
+        }
+    }
+
+    fn undo(&self, entity: &mut Option<AccountAssetActivity>) {
```
slowbackspace: 🟢 Latent trap: `new()` accepts `Vec<(Bytes, u128)>`, which can repeat a unit. For `[(U,5),(U,3)]`, `apply` snapshots `prev = [(U,orig),(U,orig+5)]`, and this forward-order restore leaves U at `orig+5` instead of `orig` — silent total corruption on rollback. Unreachable today (the producer collects from a `BTreeMap`), but either iterate `.rev()` here or make the record hold `BTreeMap<Bytes, u128>` directly — that also deletes the `collect()` conversions at the call site and makes duplicates unrepresentable.

crates/minibf/src/routes/accounts.rs
```diff
@@ -306,6 +308,66 @@ where
...
+        .read_cardano_entity::<AccountActivity>(account_key.entity_key.as_slice())?
+        .unwrap_or_default();
+
+    let assets = domain
+        .read_cardano_entity::<AccountAssetActivity>(account_key.entity_key.as_slice())?
+        .unwrap_or_default();
+
+    let stake_address = account_key
+        .address
+        .to_bech32()
```
slowbackspace, line 448: 🟢 Parity note, document-only: for a registered-but-never-funded account, BF returns `stake_address: null` (its SQL derives the echo from a `tx_out` subselect that finds no rows), while this returns the normalized bech32. The dolos behavior is the saner one; flagging it only so it's a known divergence if a fixture ever hits that case.

slowbackspace (commented):
Severity legend: 🔴 must fix · 🟡 should fix · 🟢 minor.

Reviewed with live verification on preview:

- Semantics match blockfrost-backend-ryo's ground-truth SQL, including failed-tx collateral handling (db-sync swaps in collateral flows for invalid txs exactly like pallas `consumes()`/`produces()` — verified equivalent, so no comment on that).
- The official blockfrost-tests preview fixture passes on a store rebuilt with this branch (`accounts/:stake_address/addresses/total big stake address`, strict match). A full accounts-family run shows zero regressions against a main-baseline: identical pre-existing failures, plus the new endpoint flipping to pass.
- `cargo test` (dolos-cardano, dolos-minibf, dolos-snapshot) and clippy are clean on a local merge with main.

Two housekeeping notes:

- The branch needs a rebase: trivial conflict with `GovDistrBoundaryCredit` in the `CardanoDelta` enum on main.
- The head commit message says "governance dreps endpoint - tests" but contains the snapshot/goldens changes for this feature.

slowbackspace (approved):
some nitpicks but LGTM

scarmuega:
`AccountActivity` looks like it could live on `AccountState` rather than being its own entity.

The keys are byte-identical — `AccountActivityRecord::key` and `ControlledAmountInc::key` are both `minicbor::to_vec(&cred)`. And the row sets aren't merely the same cardinality, they're nested: the activity accumulation in `AccountVisitor` sits inside `if !is_pointer`, while `ControlledAmountInc`/`Dec` are emitted unconditionally for the same credential in the same hooks, so no credential can have activity without already having an `AccountState`. The new endpoint already leans on that — it 404s on `AccountState` existing and then `unwrap_or_default()`s the activity read.

Merging would be three `#[cbor(default)]` fields at indices 8–10, the same additive pattern `retired_pool` already uses at `#[n(7)]`. It saves a second read-modify-write per tx on a key that's being written anyway, drops a duplicate row per account, and keeps one namespace out of the stele.

The things that could block it look fine: `StakeDeregistration` sets `deregistered_at` without deleting the row, so the lifetime counters survive; and `AccountVisitor` already carries `epoch`, so `flush_tx_activity` can build through `AccountState::new`.

The one real cost is `AccountTransition` — it sweeps every account at each epoch boundary and would rewrite ~40 extra bytes per row. That seems cheap next to the per-tx double write it removes, but worth weighing.

scarmuega:
Some measured numbers on what `AccountAssetActivity` costs, taken against a synced mainnet instance.

**Baseline.** Mainnet state today is 18.68M entity rows — 1.99 GB of values plus 747 MB of keys, 2.73 GB total. `accounts` is 5.92M rows / 783 MB, `assets` is 11.16M rows / 991 MB.

**Disk.** The live UTxO set holds 13.78M distinct (credential, asset) pairs across 914k credentials. `received_assets` alone encodes to 866 MB, at 62.9 bytes per pair to carry a 16-byte quantity — `[u8; 16]` derives as a CBOR array of 16 integers, 17 to 33 bytes. With `sent_assets` that is roughly 1.3 GB, a ~48% increase in state size. It is a floor: the entity stores everything ever received or sent rather than currently held, and `accounts` already carries 5.92M rows against the 2.52M credentials present in the live set.

**Processing.** The fold decodes and re-encodes the full map once per batch per touched account. The distribution is long-tailed — p50 0 assets, p99 94, p99.9 555, max 196,908. 921 credentials hold >= 1k assets, 22 hold >= 10k, 3 hold >= 50k.

| distinct assets | encoded value | decode + encode |
| --- | --- | --- |
| 10,000 | 953 KB | 2.5 ms |
| 22,083 | 2.10 MB | 6.6 ms |
| 49,676 | 4.74 MB | 16.3 ms |
| 88,958 | 8.48 MB | 34.4 ms |
| 196,908 | 18.9 MB | 72.3 ms |

Those are multi-megabyte values rewritten through every LSM compaction level, on the sync path, with the cost scaling by how often the account appears in blocks.

**Rollout.** There is no incremental path — existing nodes need a full `doctor rebuild-state`. The namespace also enters `snapshot::NAMESPACES`, so every stele carries the extra bytes and every operator bootstrapping mainnet downloads them.

For context, `tx_count` and the lovelace sums come from `AccountActivity` at ~40 bytes per account. Everything above is the per-asset breakdown alone.

michalrus:
@vladimirvolek more Git conflicts :pray:

docs/content/configuration/schema.mdx
```diff
@@ -266,7 +266,7 @@ The `sync` section controls how Dolos synchronizes the chain from upstream peers
 | sync_limit      | string  | "NoLimit" (default) |
 
 - `pull_batch_size`: the number of blocks that are fetched per batch (defaults to `100`).
-- `max_history`: maximum number of slots to keep in archive history before housekeeping prunes old data.
+- `max_history`: maximum number of slots to keep in archive history before housekeeping prunes old data. Pruned blocks are also out of reach for the MiniBF endpoints that fold lifetime totals (`/addresses/{address}/total`, `/accounts/{stake_address}/addresses/total`), whose sums then only cover the retained window.
```
scarmuega, line 269: this seems like a little too much detail for this section.

docs/content/apis/minibf.mdx
```diff
@@ -97,6 +97,8 @@ Check the [Configuration Schema](../configuration/schema) for detailed info on h
 
 Dolos provides many, but not all of the Blockfrost endpoints. The following list represent the endpoints currently supported by Dolos:
 
+Two of them report lifetime totals — `/addresses/{address}/total` and `/accounts/{stake_address}/addresses/total` — by folding every transaction the address or account ever took part in, read from the archive on each request. They are only as complete as the history the node keeps: with `sync.max_history` set, the sums and `tx_count` cover the retained window instead of the whole chain, so a node serving them should run with full history.
```
scarmuega, line 100: seems too specific for this page.

michalrus:
@vladimirvolek there are conflicts 🙏 I'll move it back to in-progress temporarily

scarmuega (approved):

PR #1202, opened 2026-08-12
fix(stelae): a staging directory that cannot be used says which one

scarmuega:
Plan: `plans/dolos-stelae-scratch-dir-diagnostics.md` (Trellis domain root), filed as residue of #1191.

## What was wrong

`--scratch-dir` on `dolos snapshot publish` and `dolos bootstrap stelae` takes a
filesystem path an operator types, and its default is derived from `storage.path`.
Both are paths people get wrong: a typo, a volume that is not mounted yet, a
directory owned by somebody else, a path that is already a regular file.

Pointed at a regular file, a publish said this and nothing else:

```
Error:   × publishing the stele
  ├─▶ stelae error: io error: File exists (os error 17)
  ├─▶ io error: File exists (os error 17)
  ╰─▶ File exists (os error 17)
```

Three lines saying the same thing three times, and not one of them names the
path, the flag, or the fact that this was a staging directory rather than the
stele, the registry, or a store. `stelae::oci::Shared::scratch` raised both of
its fallible calls — `create_dir_all` and `tempfile_in` — through
`Error::Io`'s `#[from]`, and `std::io::Error` carries the errno without the path
that produced it.

## What changed

`Error::Scratch { dir, source }` in `crates/stelae/src/lib.rs`, raised by the
scratch path for both calls that can fail against a directory somebody named.
The variant belongs in `stelae` because that is where the path is in scope — a
`miette` context on the `dolos` side would name the flag but not the path, since
the binary cannot tell which of the transport's many IO calls failed.

The message names the directory and what it was for and stops there; the
operating system's own words stay on the `#[source]`, one line further down the
chain, so a rendered report says each thing once:

```
  ├─▶ stelae error: cannot use /mnt/fast/scratch as the staging directory
  ╰─▶ File exists (os error 17)
```

The unnamed case (`scratch_dir: None`, the platform temporary directory) keeps
the catch-all — nobody chose that path, so there is nothing useful to name.

**No new check and no new refusal.** The failure already happened; only what it
says changes. Sizing the staging volume before the fact is the preflight's job
(#1192), and a check here would collide with it. The flags, their defaults and
their help are untouched (#1191).

## Tests

- `crates/stelae/src/oci.rs` — `an_unusable_staging_directory_names_itself`
  covers the scratch path directly against a path that is an existing regular
  file, which `create_dir_all` cannot turn into a directory for anybody, `root`
  included. Under plain `cargo test`: no container, no crypto provider. This is
  why `scratch` became a free function rather than staying a method on `Shared`
  — a method needs a registry client to reach.
- `crates/stelae/tests/oci.rs` —
  `a_staging_directory_that_cannot_be_used_names_itself` makes the same claim
  through a real publish and a real pull. Both directions, because one
  `Options::scratch_dir` serves the sink on the way up and the pulled blob on the
  way down, and fixing whichever direction was tested first is how this defect
  comes back. Rides the existing `registry.yml` gate; no new CI wiring.

## Verification

Run against a clean `main` worktree:

- `cargo test -p stelae --all-features` — pass
- `cargo test -p stelae --all-features --test oci -- --ignored a_staging_directory_that_cannot_be_used_names_itself` — pass against a real `registry:2`
- `cargo clippy --all-targets --all-features -- -D warnings` — clean
- `cargo +nightly fmt --all -- --check` — clean
- `cargo deny check advisories` — ok
- `cargo tree -p stelae -e normal --all-features` — no `^dolos(-|$)` package; the ADR-004 boundary holds

The full-workspace `cargo test` could not be run locally — the host disk filled
mid-build (121 MiB free of 460 GiB), which is a machine condition and not this
change. It is left to CI, which is why this is a draft.

## Note for the reviewer

The plan's done criterion 1 names a CLI test at `tests/stelae_scratch.rs`. That
file never landed on `main` — it existed on `feat/stelae-scratch-dir` and was
dropped from #1191's squash — so there was nothing to amend. The criterion's
substance (a publish and a restore each told to stage at a regular-file path
fail with a message naming the path and identifying it as the staging directory)
is covered in `crates/stelae/tests/oci.rs` instead, which is already Docker-gated
and already wired into `registry.yml`. Recreating the CLI suite would also mean a
new `registry.yml` step and its `AGENTS.md` mirror — scope the plan did not
budget, and adjacent to the flag work it explicitly fences off. Raised as an open
escalation on the plan for `org/founder`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error reporting when temporary staging directories cannot be created or used.
  * Errors now identify the affected directory and provide clearer staging context.
  * Preserved existing behavior for unnamed temporary files.

* **Tests**
  * Added coverage for staging failures during uploads and downloads.
  * Added validation for successful unnamed temporary-file staging.

### DIFF
--- a/crates/stelae/src/lib.rs
+++ b/crates/stelae/src/lib.rs
@@ -283,6 +283,29 @@ pub enum Error {
     #[error("{value:?} is not an OCI repository: {reason}")]
     InvalidRepository { value: String, reason: String },
 
+    /// A staging directory the transport could not use.
+    ///
+    /// Distinct from [`Error::Io`], which is the catch-all every other bare
+    /// `?` in this crate falls through, because this is the one path an
+    /// operator types on a command line — `dolos snapshot publish
+    /// --scratch-dir`, `dolos bootstrap stelae --scratch-dir`, or the default
+    /// the binary derives from `storage.path`. A typo, an unmounted volume, a
+    /// directory owned by somebody else and a path that is already a regular
+    /// file all arrive here, and `std::io::Error` carries none of them: it
+    /// knows the errno and not the path that produced it. So the path is
+    /// captured where it is still in scope, which is the only place it can be.
+    ///
+    /// The message deliberately does *not* repeat what the operating system
+    /// said — that stays on the source, one line further down the chain, so a
+    /// rendered report says each thing once.
+    #[cfg(feature = "oci")]
+    #[error("cannot use {} as the staging directory", dir.display())]
+    Scratch {
+        dir: std::path::PathBuf,
+        #[source]
+        source: std::io::Error,
+    },
+
     /// Anything the registry client reported.
     #[cfg(feature = "oci")]
     #[error("registry error: {0}")]
--- a/crates/stelae/src/oci.rs
+++ b/crates/stelae/src/oci.rs
@@ -735,11 +735,9 @@ fn is_absent(error: &oci_client::errors::OciDistributionError) -> bool {
 /// A staged file, created where [`Options::scratch_dir`] said.
 ///
 /// Both calls that can fail against a *named* directory raise
-/// [`Error::Scratch`] rather than falling through `#[from]` into
-/// [`Error::Io`]: this directory came from an operator — `--scratch-dir`, or
-/// the default a host derives for them — and an errno with no path in it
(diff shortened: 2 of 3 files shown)


scarmuega:
@coderabbitai review

PR #1220, opened 2026-08-17
feat(minibf): add `/assets/{subject}/txs` endpoint

### LINKED ISSUE #1100: minibf: add `/assets/{asset}/txs`
Implement the Blockfrost `/assets/<asset>/txs` endpoint in the `minibf` crate.

### Effort

S — Low effort (pure mapping)

### Approach

Hash-only subset of existing `/assets/<asset>/transactions`

### Why now

0.0020% of observed traffic.

vladimirvolek:
resolves: https://github.com/txpipe/dolos/issues/1100

## Summary by CodeRabbit

* **New Features**
  * Added an asset transaction-hash lookup endpoint with pagination and preserved ordering.
  * Added support for retired-pool listings and pool certificate-update lookups.

* **Bug Fixes**
  * Improved asset-subject validation across asset and address queries.
  * Unknown but valid assets now return clear “Not Found” responses.
  * Invalid asset inputs receive consistent error handling.

* **Documentation**
  * Updated API documentation for the newly supported endpoints and clarified pool and epoch descriptions.

### DIFF
--- a/crates/minibf/src/routes/assets.rs
+++ b/crates/minibf/src/routes/assets.rs
@@ -681,11 +681,18 @@ pub async fn by_subject_transactions<D>(
 ) -> Result<Json<Vec<AssetTransactionsInner>>, Error>
 where
     D: Domain + Clone + Send + Sync + 'static,
+    Option<AssetState>: From<D::Entity>,
 {
     let pagination = Pagination::try_from(params)?;
     pagination.enforce_max_scan_limit(domain.config.max_scan_items())?;
 
     let subject = hex::decode(&subject).map_err(|_| Error::InvalidAsset)?;
+
+    // Blockfrost returns 404 for a valid but unknown asset, same as `/addresses`.
+    let entity_key = pallas::crypto::hash::Hasher::<256>::hash(subject.as_slice());
+    if !domain.cardano_entity_exists::<AssetState>(entity_key.as_slice())? {
+        return Err(StatusCode::NOT_FOUND.into());
+    }
 
     let (start_slot, end_slot) = pagination.start_and_end_slots(&domain).await?;
     let stream = domain.query().blocks_by_asset_stream(
@@ -727,6 +734,24 @@ where
         .collect();
 
     Ok(Json(transactions))
+}
+
+/// Alias of `/assets/{subject}/transactions` that returns only the tx hashes.
+/// Same scan, same pagination, thinner payload.
+pub async fn by_subject_txs<D>(
+    path: Path<String>,
+    params: Query<PaginationParameters>,
+    state: State<Facade<D>>,
+) -> Result<Json<Vec<String>>, Error>
+where
+    D: Domain + Clone + Send + Sync + 'static,
+    Option<AssetState>: From<D::Entity>,
+{
+    let Json(transactions) = by_subject_transactions(path, params, state).await?;
+
+    let hashes = transactions.into_iter().map(|x| x.tx_hash).collect();
+
+    Ok(Json(hashes))
 }
 
 fn collect_minted_subjects(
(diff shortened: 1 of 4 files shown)


crates/minibf/src/routes/assets.rs
```diff
@@ -681,12 +681,19 @@ pub async fn by_subject_transactions<D>(
 ) -> Result<Json<Vec<AssetTransactionsInner>>, Error>
 where
     D: Domain + Clone + Send + Sync + 'static,
+    Option<AssetState>: From<D::Entity>,
 {
     let pagination = Pagination::try_from(params)?;
     pagination.enforce_max_scan_limit(domain.config.max_scan_items())?;
 
     let subject = hex::decode(&subject).map_err(|_| Error::InvalidAsset)?;
```
slowbackspace: 🟡 Blockfrost validates the asset length before the lookup. `validateAsset` (ryo) and `validate_asset_name` (mimicry, `src/asset.rs:32`) require 56–120 hex chars and return 400 otherwise. Dolos only runs `hex::decode`. With the new 404 gate, `GET /assets/abcd/transactions` now returns 404 where Blockfrost returns 400 'Invalid or malformed asset format.' (verified live against this branch). Add a length check that maps to 400. The same gap exists in `by_subject` and `by_subject_addresses`; a shared helper would fix all of them.

crates/minibf/src/routes/assets.rs
```diff
@@ -681,12 +681,19 @@ pub async fn by_subject_transactions<D>(
...
 
+    // Blockfrost returns 404 for a valid but unknown asset, same as `/addresses`.
```
slowbackspace: 🟢 The comment cites the wrong precedent. `/addresses/{address}/transactions` checks existence lazily, only when the page is empty (`addresses.rs:550-556`), and the `/addresses/{address}/txs` alias has no check at all. The matching precedent is `/assets/{subject}/addresses` in this file. Suggest: name that endpoint, or state the Blockfrost rule without a cross-reference.

crates/minibf/src/routes/assets.rs
```diff
@@ -681,12 +681,19 @@ pub async fn by_subject_transactions<D>(
...
+    let entity_key = pallas::crypto::hash::Hasher::<256>::hash(subject.as_slice());
+    if !domain.cardano_entity_exists::<AssetState>(entity_key.as_slice())? {
```
slowbackspace: 🟡 This is the third inline copy of the decode → `Hasher::<256>` → `cardano_entity_exists` gate in this file (`by_subject` line 484, `by_subject_addresses` line 525). The copies already drift: `by_subject` maps bad hex to `StatusCode::BAD_REQUEST` and uses `read_cardano_entity`. Extract one helper that decodes the subject, validates it, checks existence, and returns the subject bytes. Call it from all three handlers. That also gives the 56–120 length check a single home.

crates/minibf/src/routes/assets.rs
```diff
@@ -729,6 +736,24 @@ where
...
+pub async fn by_subject_txs<D>(
+    path: Path<String>,
+    params: Query<PaginationParameters>,
+    state: State<Facade<D>>,
+) -> Result<Json<Vec<String>>, Error>
+where
+    D: Domain + Clone + Send + Sync + 'static,
+    Option<AssetState>: From<D::Entity>,
+{
+    let Json(transactions) = by_subject_transactions(path, params, state).await?;
```
slowbackspace: 🟡 The alias inherits `from`/`to` through `PaginationParameters`, but Blockfrost's `/assets/{asset}/txs` does not define them. Ryo's handler reads only order/count/page (no `getAdditionalParametersFromRequest`). Mimicry parses `from`/`to` but its `assets_asset_txs.sql` binds only order, count, page, and asset — the values never reach the query. So both references return the full list for `GET /assets/{unit}/txs?from=X&to=Y`, while this alias filters by the range. Suggest clearing `from`/`to` from the params before delegating.

slowbackspace (changes requested):
Reviewed the code and verified behavior against ryo, mimicry, and the official blockfrost-tests suite.

**Severity legend:** 🔴 must fix · 🟡 should fix · 🟢 optional

**Verification done:**
- Ran blockfrost-tests on preview with this branch: `assets/:asset/txs` 5 passed, `assets/:asset/transactions` 9 passed.
- Ran blockfrost-tests on mainnet: happy-path, ordering, and from/to tests pass on both endpoints. Six deep-pagination tests fail (`page=42420`, `page=10000`) with the `max_scan_items` 400. The same tests fail on `/transactions` before this PR. This is inherited, not a regression.
- The entity-key hashing matches write-time (`MintStatsUpdate::key`). The eager 404 pre-check matches both ryo (`assets_404`) and mimicry (`is_asset_on_chain`).

One request for the PR description: this PR changes the contract of the shipped `/assets/{subject}/transactions` endpoint. A well-formed but unknown asset returned `200 []` before. It now returns `404`. The change is correct Blockfrost parity, but please call it out as a behavior change, not only a feature addition.

vladimirvolek:
@copilot resolve the merge conflicts in this pull request

scarmuega (approved):

PR #1222, opened 2026-08-18
feat(doctor): rebuild-state — regenerate the state store from the local archive

scarmuega:
## Plan

Implements [`dolos-state-rebuild-from-archive`](https://github.com/scarmuega/brain-txpipe) (Trellis plan, org/coder): an operator debugging or repairing a synced dolos instance can regenerate its state store from the instance's own archive — no network, no snapshot re-import, no re-writing of the archive or indexes.

## What changed

- **New `dolos doctor rebuild-state`** (`src/bin/dolos/doctor/rebuild_state.rs`): replays the instance's archive from origin through `import_blocks` into a fresh state store, against a hand-assembled `DomainAdapter` (in-memory WAL, no-op archive/indexes, ephemeral mempool).
  - **In place (default):** crash-safe sequence — preflight → `wal.reset_to(Origin)` → wipe `state_path` → replay to archive tip → `wal.reset_to(final cursor)` → postflight. A crash mid-rebuild leaves WAL(origin) behind the state cursor, which the next startup refuses loudly (`InconsistentState`); re-running the command recovers.
  - **`--target <path>` / `--ephemeral`:** isolated outputs; the instance's stores are not touched.
  - **`--stop-epoch <n>`** (isolated outputs only), **`--chunk <n>`**, **`--force`** (non-TTY requires it; TTY confirms the wipe interactively).
  - **Preflight** refuses empty and pruned (`max_history`) archives; deliberately *not* a strict prev-hash origin walk (Byron EBB slot-key overwrite would flag legitimate mainnet archives). Continuity is enforced during replay by `check_extension`.
- **`--rewrite-logs` (in-place only):** the domain's archive is swapped for a new write-gated `ArchiveStoreBackend::LogsOnly` variant wrapping the *already-open* redb store (`src/adapters/storage.rs`): `write_log`/`commit` delegate, `apply`/`undo` no-op (block appends are not idempotent). Boundary log keys are slot-derived, so corrected StakeLog / reward logs / EpochState rows overwrite in place. Overwrites only; never deletes stale rows.
- **Rebuild-equality test** (`tests/rebuild_state.rs`): a synthetic chain built through import, state rebuilt in place from the archive by the real binary, full store equality asserted (cursor, every namespace's entities, full UTxO set), archive segment files byte-identical, WAL reseeded; plus `--rewrite-logs` row-equality, `--target`/`--ephemeral` isolation, non-TTY `--force` refusal, and the `--stop-epoch` flag gate. The `Node` fixture now persists the synthetic chain's `custom_utxos` into `dolos.toml` so a separate process replaying from origin seeds the same genesis.

## Scope notes (from the plan)

- Always from origin; no seeded mid-chain starts (#1018).
- Epoch-test adoption of the archive block source is a separate follow-up plan.
- Byron block-count drift in rebuilt `RollingStats` (one per Byron epoch, from the EBB slot-key overwrite) is documented, not fixed here.

### Automated

- `cargo test` — all suites pass, including the new `rebuild_state` suite (also under `--features strict`, per done criterion 6)
- `cargo clippy --all-targets --all-features -- -D warnings` — clean
- `cargo +nightly fmt --all -- --check` — clean

### Preprod, end to end

A preprod instance bootstrapped from Mithril (`~/dolos-instances/preprod-rebuild`, 17 GB archive / 304 segments, tip slot 131284793), then driven through every mode. Wall-clock on a 10-core / 16 GB workstation:

| run | wall-clock |
|---|---|
| Mithril bootstrap (download + full-pipeline import) — **the datum** | **93.8 min** |
| in-place `rebuild-state` | **42.4 min** |
| `--rewrite-logs` | 49.0 min |
| `--target` (fresh state store elsewhere) | 43.6 min |
| recovery rebuild after a killed run | 42.6 min |
| `dolos data check` (full width) | 1.5 min |

The plan's reference datum was "~1 h preprod" for a full replay; the measured bootstrap here was 93.8 min including download. A state rebuild costs **42.4 min and re-writes nothing but the state store** — no download, no snapshot re-import, no archive or index rewriting.

**Criterion 1 — in-place rebuild.** Regenerated state in place; cursor returned to the synced tip `131284793(9689ca2a…)`; all 304 archive segment files byte-identical (sha256) before/after; `dolos data check` passes at full width (cursors, archive-continuity, account-epochs, epoch-log, totals — 0 issues); `dolos daemon` then bootstraps cleanly ("WAL is in sync with state"), finds intersection at the rebuilt tip and rolls forward onto live blocks with zero errors.

**Criterion 2 — `--rewrite-logs`.** Derived log rows re-written and spot-checked via `dolos data dump-logs` (`stakes`, `epochs` namespaces) — identical values at identical keys; blocks table and segment files unchanged in count, size, and bytes.

**Criterion 3 — isolated outputs.** `--target` completed with the instance's stores untouched (log rows unchanged, segment hashes unchanged). `--ephemeral` completed writing nothing.

> ⚠️ **Operational limit found:** an *unbounded* `--ephemeral` replay of the full preprod chain exhausted the 16 GB workstation and took the machine down. The builtin memory state store holds every entity and the whole UTxO set uncompressed, which is a different order of magnitude from the same state in fjall's on-disk LSM tree. Documented on the flag and in the module docs (2nd commit); `--ephemeral` is for bounded replays (`--stop-epoch`) and small chains, `--target` is the full-chain validation mode. The instance was verified untouched afterwards — a crash during an ephemeral run costs nothing.

**Criterion 4 — `--stop-epoch`.** Stops cleanly at the requested epoch, verified as a scaling curve with `--target`:

| stop epoch | wall-clock | state size | cursor | exact boundary |
|---|---|---|---|---|
| 20 | 2.9 s | 4.8 M | 6,998,401 | 6,998,400 |
| 30 | 4 s | 15 M | 11,318,425 | 11,318,400 |
| 60 | 50 s | 302 M | 24,278,405 | 24,278,400 |
| 100 | 129 s | 786 M | 41,558,400 | 41,558,400 |

Every cursor lands within one block of the exact boundary, and cost grows superlinearly toward the full-chain figures — preprod's first ~30 epochs are genuinely near-empty.

**Criterion 5 — crash recovery.** An in-place rebuild killed 90 s into the replay (`kill -9`) leaves the instance refusing to start, legibly:

```
Error:   × state (slot 34253401) is ahead of WAL (slot 0)
  help: run `dolos doctor reset-wal` to rebuild the WAL from the current state
```

Re-running `dolos doctor rebuild-state --force` recovers it fully: same cursor as the original sync, `data check` clean at full width, daemon starts.

> 📝 **Finding, not fixed here (pre-existing, out of this PR's scope):** that `help:` line points at `doctor reset-wal`, which is the wrong advice for a *half-rebuilt* instance — it would seed the WAL at the partial cursor and present a truncated ledger as consistent, instead of finishing the rebuild. The refusal itself is correct and is what the plan's crash-safety story relies on; only the suggested remedy is misleading on this path. Flagged for the owner to decide, since the message is shared by every crash path, not just this command's.

## Summary by CodeRabbit

* **New Features**
  * Added `dolos doctor rebuild-state` to regenerate state from the local archive.
  * Supports in-place, alternate-directory, and temporary in-memory rebuilds.
  * Added bounded replay, configurable chunking, log rewriting, and forced execution options.
  * Archive statistics, export sanitization, and pruning now support logs-only storage.

* **Bug Fixes**
  * Improved rebuild safety through validation, recovery handling, and completion checks.

* **Tests**
  * Added end-to-end coverage for rebuild modes, log rewriting, safety checks, and archive integrity.

### DIFF
--- a/src/adapters/storage.rs
+++ b/src/adapters/storage.rs
@@ -668,6 +668,12 @@ impl ArchiveStoreBackend {
     /// Clones the handle out of the open store rather than opening the path
     /// again, because redb refuses to open the same file twice. Returns
     /// `None` when there is no redb store to wrap.
+    ///
+    /// The result therefore *aliases* the store it came from: both share one
+    /// `Arc<Database>`. Anything reaching for exclusive database access —
+    /// `db_mut`, and so redb compaction — must refuse a `LogsOnly` value
+    /// rather than treat it as a `Redb` one, because `Arc::get_mut` cannot
+    /// succeed while the original handle is alive.
     pub fn logs_only(&self) -> Option<Self> {
         match self {
             Self::Redb(s) | Self::LogsOnly(s) => Some(Self::LogsOnly(s.clone())),
--- a/src/adapters/storage.rs
+++ b/src/adapters/storage.rs
@@ -632,6 +632,16 @@ impl CoreStateStore for StateStoreBackend {
 #[derive(Clone)]
 pub enum ArchiveStoreBackend {
     Redb(dolos_redb3::archive::ArchiveStore),
+    /// Write-gated view over an already-open redb archive: reads and
+    /// derived-log writes pass through, block appends and undos are
+    /// discarded.
+    ///
+    /// This exists for replays over an archive that already holds the chain
+    /// (`dolos doctor rebuild-state --rewrite-logs`): block appends are not
+    /// idempotent — replaying them would double every segment file — while
+    /// boundary log keys are slot-derived and identical under replay, so
+    /// re-written log rows overwrite the originals in place.
+    LogsOnly(dolos_redb3::archive::ArchiveStore),
     NoOp(NoOpArchiveStore),
 }
 
@@ -650,6 +660,19 @@ impl ArchiveStoreBackend {
     /// Create a no-op archive store that discards all writes.
     pub fn noop() -> Self {
         Self::NoOp(NoOpArchiveStore)
+    }
+
+    /// Wrap this backend's already-open redb store in a [`Self::LogsOnly`]
+    /// write gate.
+    ///
+    /// Clones the handle out of the open store rather than opening the path
+    /// again, because redb refuses to open the same file twice. Returns
+    /// `None` when there is no redb store to wrap.
+    pub fn logs_only(&self) -> Option<Self> {
+        match self {
+            Self::Redb(s) | Self::LogsOnly(s) => Some(Self::LogsOnly(s.clone())),
+            Self::NoOp(_) => None,
+        }
     }
 
     /// Create an in-memory archive store.
@@ -678,7 +701,7 @@ impl ArchiveStoreBackend {
 
     pub fn shutdown(&self) -> Result<(), ArchiveError> {
         match self {
-            Self::Redb(s) => s
+            Self::Redb(s) | Self::LogsOnly(s) => s
                 .shutdown()
                 .map_err(|e| ArchiveError::InternalError(e.to_string())),
             Self::NoOp(s) => s.shutdown(),
@@ -688,6 +711,8 @@ impl ArchiveStoreBackend {
 
 pub enum ArchiveWriterBackend {
     Redb(Box<<dolos_redb3::archive::ArchiveStore as CoreArchiveStore>::Writer>),
+    /// Delegates `write_log` and `commit`; discards `apply` and `undo`.
+    LogsOnly(Box<<dolos_redb3::archive::ArchiveStore as CoreArchiveStore>::Writer>),
     NoOp(NoOpArchiveWriter),
 }
 
@@ -695,6 +720,7 @@ impl CoreArchiveWriter for ArchiveWriterBackend {
     fn apply(&self, point: &ChainPoint, block: &RawBlock) -> Result<(), ArchiveError> {
         match self {
             Self::Redb(w) => w.apply(point, block),
+            Self::LogsOnly(_) => Ok(()),
             Self::NoOp(w) => w.apply(point, block),
         }
     }
@@ -706,7 +732,7 @@ impl CoreArchiveWriter for ArchiveWriterBackend {
         value: &EntityValue,
     ) -> Result<(), ArchiveError> {
         match self {
-            Self::Redb(w) => w.write_log(ns, key, value),
+            Self::Redb(w) | Self::LogsOnly(w) => w.write_log(ns, key, value),
             Self::NoOp(w) => w.write_log(ns, key, value),
         }
     }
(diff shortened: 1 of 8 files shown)


scarmuega:
Addressed the two actionable review findings plus the nitpick in 9e55f511.

**1. `--chunk 0` destroyed the instance (confirmed, fixed).** Valid and the most serious of the three: with `--chunk 0` the replay's `take(0)` returned an empty batch immediately, so the command bailed on the missing cursor — in in-place mode that lands *after* the WAL reset and the state wipe, so a one-character typo left an unusable instance. Now refused up front, before anything destructive, with a regression test (`a_zero_chunk_is_refused_before_touching_anything`) asserting the instance is byte-for-byte untouched after the refusal.

**2. `LogsOnly` reaching `db_mut()` (confirmed, fixed).** Also valid, and worse than "no live caller reaches it": `logs_only()` exists precisely to clone a handle while the original stays alive, so a `LogsOnly` value *always* aliases the `Arc<Database>` — `Arc::get_mut(..).unwrap()` in `db_mut` could never succeed, not merely "whenever a second handle exists". Folding it into the `Redb` arm of `prune-chain` and `export` made a guaranteed panic look like a supported path. Both now refuse it explicitly with an operator-facing message, and the aliasing contract is documented on `logs_only()` itself.

**3. `expect` on the archive tip (nitpick, taken).** Swapped for a `bail!`, matching the operator-facing error style used everywhere else in the command.

Verification after the fixes: `cargo test` (full workspace, including the 6-test `rebuild_state` suite), `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo +nightly fmt --all -- --check` all clean locally.

---

**On the failing `cargo deny` check — not from this PR, and it is not the yanked crate.**

The job's tail shows `spin 0.9.8 — yanked version`, but `deny.toml` already sets `yanked = "warn"`, so that is not what fails the gate. The actual error is:

```
error[vulnerability]: h2 unbounded empty DATA frames
  ID: RUSTSEC-2026-0258   (Low severity)
```

a newly-published advisory against `h2`, reached via `hyper` / `reqwest` / `tonic`. Evidence it is pre-existing and repo-wide:

- this PR changes no dependencies at all — `git diff main...HEAD -- Cargo.toml Cargo.lock crates/*/Cargo.toml` is empty
- `h2` resolves to the same `0.3.26` / `0.4.9` on `main` as on this branch
- `main`'s own CI run today fails the same `cargo deny (advisories, bans)` job, as do several unrelated open branches

So it wants a lockfile bump on `main` (fixing `main` and every open PR at once) rather than a dependency change smuggled into a `doctor` subcommand PR. Happy to open that as a separate PR if wanted.

PR #1225, opened 2026-08-18
feat(indexes): stake address log for O(page) account address queries

slowbackspace:
Stacked on #1221 (the base branch). Follow-up to the performance concern flagged there.

## Problem

With the correct first-appearance ordering from #1221, `GET /accounts/{stake_address}/addresses?order=desc` must scan the account's entire archive history before it can reverse the list. Measured on a fully synced mainnet snapshot: **305 seconds** (cold and warm — the cost is CPU-bound block decoding) and ~274MB transient RSS for an exchange account with 400k+ addresses. Any `desc` page on such an account pays the full scan, which is also a DoS shape on public nodes.

## Design

A **stake address log**: each `(stake credential, address)` pair stored once, at its first on-chain appearance, ordered by `(slot, tx order, output order)`. Both `asc` and `desc` become a single page read (`desc` is a reverse range read — Blockfrost defines it as the exact reverse of `asc`).

- **fjall** (the supported index backend) gets a new `stake-log` keyspace with two entry shapes: a membership entry per pair (the write-path probe and the undo key) and an ordered entry (the page read). The write batch cannot read its own pending inserts, so the writer keeps a batch-local seen-set — one writer can span many blocks in WAL catch-up.
- **memory** implements the same semantics, so `ToyDomain` endpoint tests exercise the log. **redb3** and **noop** answer `None` (redb3 is deprecated for index stores).
- **Undo correctness**: rollback does not re-run `index_block`, so `compute_undo` derives the appearances from the same `stake_appearances_from_block` function the apply path uses. A pair is removed only when the undone block *is* its stored first appearance; a block that merely repeated an address leaves the log untouched.
- **Rollout**: a ready marker gates reads. Genesis bootstrap sets it, so any store synced (or mithril-bootstrapped) from scratch serves the log; existing stores answer `None` and the endpoint transparently falls back to the archive scan from #1221 until they resync. No doctor/backfill job — resync is the rollout path.
- Full address bytes are stored (not just the payment part) so pointer addresses round-trip. Estimated mainnet size: 5–15GB against a 340GB store.

## Testing

- Backend conformance test in `tests/index_roundtrip.rs` runs against fjall and memory: marker gating, first-appearance dedup (including inside one multi-block writer), ordered paging from both ends, offset windows, stake isolation, and undo-of-first-appearance-only.
- Memory-store unit tests cover the same semantics at the store level.
- Endpoint equivalence test: two identical synthetic chains, one serving from the log and one from the archive-scan fallback, must return identical responses for every query shape. Verified non-vacuous by flipping the order flag and watching the order tests fail.
- Full workspace suite passes; clippy clean; nightly fmt applied.

Not yet done: an end-to-end run against a freshly synced network store (my local preview/mainnet stores predate the log, so they exercise only the fallback). Happy to re-sync preview and attach numbers if you want them before undrafting.

### DIFF
--- a/crates/cardano/src/genesis/mod.rs
+++ b/crates/cardano/src/genesis/mod.rs
@@ -170,6 +170,11 @@ pub fn bootstrap_utxos<D: Domain>(
     state_writer.commit()?;
     index_writer.commit()?;
 
+    // A genesis bootstrap only happens on a fresh store, so every block the
+    // store will ever index flows through the apply path from here on. That
+    // makes the stake address log complete by construction.
+    indexes.mark_stake_log_ready()?;
+
     Ok(())
 }
 
--- a/crates/cardano/src/indexes/delta.rs
+++ b/crates/cardano/src/indexes/delta.rs
@@ -4,8 +4,8 @@
 //! `IndexDelta` structures from Cardano block data.
 
 use dolos_core::{
-    ArchiveIndexDelta, BlockSlot, ChainPoint, EraCbor, IndexDelta, Tag, TxoRef, UtxoIndexDelta,
-    UtxoSetDelta,
+    ArchiveIndexDelta, BlockSlot, ChainPoint, EraCbor, IndexDelta, StakeAddressAppearance, Tag,
+    TxoRef, UtxoIndexDelta, UtxoSetDelta,
 };
 use pallas::{
     codec::minicbor,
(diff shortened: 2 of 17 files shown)


slowbackspace:
Applied into #1221 as commit 568d3ba4. That PR now carries both the fixes and the stake address log.

PR #1256, opened 2026-08-22
experiment: the archive index on fjall — benchmark vehicle

scarmuega:
**Not for merge.** This branch exists to produce four numbers: archive-index size, replay wall-clock, and read p50/p95 on the three real access shapes, fjall vs optimized redb, per the archive-backend evaluation in the org brain (`plans/dolos-archive-fjall-backend.md`). The verdict lands as a decision record; if it is *adopt*, a separate code-change-request productionizes for the v1.8+ line.

## What's here

- **`crates/fjall/src/archive/`** — the archive store on fjall. Two keyspaces: `archive-blocks` keeps the redb blocks table's value encoding byte-for-byte (slot → packed 16-byte locations, newest first, Byron EBB multi-block case included); `archive-logs` holds all log namespaces as `[ns_hash:8][log_key:40]`, the state store's unified-keyspace pattern. Block bodies stay in the flat segment files, reused from `dolos-redb3` as-is. No log-batch shuffle: that works around redb's half-split of ascending B-tree leaves, and an LSM memtable sorts its batch regardless.
- **`ArchiveStoreConfig::Fjall`** + adapter fan-out. `prune-chain` works through the trait; `export`, `cardinality-stats`, `rebuild-state --rewrite-logs` refuse the prototype with clear messages.
- **`tests/archive_conformance.rs`** — the redb archive tests rewritten against the trait surface and run against *both* backends, plus log semantics (roundtrip, isolation, range bounds, last-write-wins, prune/truncate temporal boundaries). redb pins the semantics; the port doesn't grade its own homework. `ToyStores` gains the archive as a third bound store, so the snapshot determinism/restore suites exercise the fjall archive through the canonical harness.
- **`dolos data bench-archive`** (hidden) — measures the three read shapes plus per-keyspace disk footprint, config-driven so both backends run identical code, seeded so both measure the same keys. Block point reads are split into slot→location resolution (the only backend-differing component) and full body read, and `FlatFileStore` now meters cumulative append/fsync time — block bodies are identical shared cost on the same disk, big enough to bury an index-path difference inside totals.

## Productionization deltas (out of scope here, listed for the follow-up)

- Relocate `flatfiles` out of `dolos-redb3` instead of coupling the crates (it is std+tempfile only).
- Crash-consistency review of the fjall commit path (currently mirrors redb's window: bodies fsynced before the index commit).
- `doctor`/`export`/stelae-restore write paths for the fjall archive.
- `data stats` footprint reporting for LSM keyspaces.

---

## Results (benchmark complete, 2026-08-23)

Sequential solo replays of the `preprod-20260814` mithril snapshot (~308 epochs), same machine, machine idle, binary `94cf84a3`. Correctness gate passed before any number was recorded: conformance suite green on both backends, `dump-logs` byte-identical per namespace (`account-epochs`, `stakes`, `epochs`) across the two replays.

| measurement | redb (optimized) | fjall | outcome |
|---|---|---|---|
| archive-index size (settled) | 906 MB allocated / 1,174 MB apparent | 468 MB (blocks 128 + logs 341) | **−48.3%** (criterion ≥30% ✓) |
| mainnet extrapolation | ~68.7 GB | ~36.1 GB | −47.4% |
| replay wall-clock | 9,309 s | 9,695 s | **1.041×** (≤1.25× ✓); index-attributable ≈1.10× after subtracting 5,432 s metered flatfile time |
| `get_block_by_slot` full path, warm p95 | 11.1 µs | 17.4 µs | 1.57× (≤2× ✓); slot→location split 3.77× warm, ~3 µs absolute |
| `by_stake_rewards` per-read p95 cold/warm | 1,365 µs / 1.4 µs | 124 µs / 2.0 µs | 0.09× / 1.44× ✓ (cold per-account 6× faster) |
| `/epochs/{n}/stakes` per-epoch p95 cold/warm | 28.5 ms / 1.7 ms | 4.6 ms / 3.5 ms | 0.16× cold ✓ / 2.04× warm (grazes; cold 6× faster) |

**Verdict: adopt** — recorded as decision 0029 in the org brain; productionization is a follow-up code-change-request in the v1.8+ line. This PR stays a draft and does not merge.

Identical seeded populations on both sides: 772 block slots, 7,937 reward hits, 156,293 scanned rows. Compression contributes only 16.3% of the size win (fjall 3.1 defaults leave L0/L1 uncompressed); the rest is structural vs redb's ~62% leaf fill, hence network-independent and safe to extrapolate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1529,6 +1529,7 @@ version = "1.7.0-alpha.0"
 dependencies = [
  "bincode 1.3.3",
  "dolos-core",
+ "dolos-redb3",
  "dolos-testing",
  "fjall",
  "pallas",
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -352,11 +352,60 @@ impl RedbArchiveConfig {
     }
 }
 
+/// Configuration for the Fjall archive backend.
+#[derive(Serialize, Deserialize, Clone, Debug, Default)]
+pub struct FjallArchiveConfig {
+    /// Optional path override for the archive directory.
+    /// If relative, resolved from storage root.
+    /// If not specified, defaults to `<storage.path>/archive`.
+    #[serde(default)]
+    pub path: Option<PathBuf>,
+    /// Optional path override for block segment files.
+    /// If not specified, segment files are stored in the archive directory.
+    #[serde(default)]
+    pub blocks_path: Option<PathBuf>,
+    /// Size (in MB) of memory allocated for caching.
+    #[serde(default)]
+    pub cache: Option<usize>,
+    /// Maximum journal size in MB.
+    #[serde(default)]
+    pub max_journal_size: Option<usize>,
+    /// Flush journal after each commit.
+    #[serde(default)]
+    pub flush_on_commit: Option<bool>,
+    /// L0 compaction threshold (default: 4, lower = more aggressive).
+    #[serde(default)]
+    pub l0_threshold: Option<u8>,
+    /// Number of background compaction worker threads.
+    #[serde(default)]
+    pub worker_threads: Option<usize>,
+    /// Memtable size in MB before flush (default: 64).
+    #[serde(default)]
+    pub memtable_size_mb: Option<usize>,
+}
+
+impl FjallArchiveConfig {
+    pub fn is_default(&self) -> bool {
+        self.path.is_none()
+            && self.blocks_path.is_none()
+            && self.cache.is_none()
+            && self.max_journal_size.is_none()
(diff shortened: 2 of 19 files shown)


scarmuega:
Superseded by #1257: the adoption verdict is in (adopt, all three criteria passed) and the mergeable feature carries the prototype forward — flatfiles extracted to a leaf crate, fjall as the default archive backend with redb supported via explicit config, conformance + restore coverage, and the benchmark reborn as divan benches. This branch remains the record of the evaluation runs.

PR #1260, opened 2026-08-24
perf(stelae): move a publish's layer round trips concurrently

scarmuega:
Plan: [`plans/dolos-stelae-publish-path-concurrency.md`](https://github.com/txpipe/brain/blob/main/plans/dolos-stelae-publish-path-concurrency.md) (Trellis domain, private)

## Why

The mainnet backfill produced the first clean measurement of the
`dolos snapshot backfill` publish path, and it contradicts the CPU-ceiling
framing the substrate was chosen for:

- A Byron epoch replays in **~10 s** on 8 vCPU. The rest of the ~6–7 min cycle
  is the **publish**, during which CPU and bandwidth both sit idle (~1 Mbps
  effective on a 15 Gbps node).
- One publish moves ~79 new layers ≈ 41 MiB — average layer ~0.5 MiB — as a
  strictly serial sequence of OCI round trips: a `HEAD` per layer of the full
  manifest, plus a create-session `POST`, chunked `PATCH`es and a commit `PUT`
  per new layer. Hundreds of sequential HTTPS round trips per epoch.
- Cycle time grows **linearly with history**: ~285 s at epoch 5 → ~428 s at
  epoch 53, a stable **+3 s/epoch** — the stele gains ~3 layers per epoch and
  every publish paid a round trip per layer of history behind it. Extrapolated
  over mainnet's remaining ~590 epochs that is **≈9 days of publish time
  alone**, crossing the backfill's abort line from the latency term rather
  than the compute term.

Nothing about that is a bandwidth problem, so the answer is not bigger layers
— the cut geometry is signed ground (decision 0025) and untouched here.

## What changed

**Concurrent layer round trips** (`crates/stelae/src/oci.rs`). A layer's round
trips are independent of every other layer's — a blob is addressed by its own
content, and no blob's upload observes another's — so they are deferred onto
the transport's runtime and run concurrently, bounded by
`Options::concurrency` (default **8**, `1` restores the serial path). The
runtime becomes multi-threaded with one worker per permit, because
`blob_stream` reads its staging file with a blocking read and a worker driving
one upload must not starve the others.

`RecordSink::finish` now closes the layer and hands the round trips to the
pool: the descriptor is a fact about bytes the sink already has, so the caller
goes back to reading its store instead of waiting on a socket.

**The manifest is the join.** `SteleWriter::seal` awaits every deferred round
trip *before the manifest is built* — not merely before it is pushed — so no
document this transport writes can name a blob the registry has not committed.
The rest of the seal's order is unchanged: config blob, sequence tag, then the
moving tag.

**A carried-forward layer costs no round trip.** `Registry::adopt_layer` takes
its blob on the word of the manifest it came from: that manifest is live in
this repository under a tag, and a registry may not reclaim a blob in that
position — one that does has already broken the stele the descriptor came
from, which the `HEAD` would not have saved either. This was the
history-proportional term. `Options::verify_adopted` /
`--verify-carried` restores the proof for an operator who does not trust their
repository's retention, and pays for it concurrently rather than in sequence.

**An upload's failure is permanent.** The join empties the handles it awaited,
so a transport that forgot would find nothing outstanding on a second seal,
agree that every layer was up, and publish a manifest naming a blob that never
landed — the one document this transport must never write, and exactly the
document concurrency makes reachable. It is remembered instead:
`Error::LayerNotWritten`, and the recovery is another publish rather than
another seal.

**Two consequences of the above, handled here rather than left to be found:**

- The byte progress bar becomes a **running total over the run** rather than a
  bar per blob. Several blobs are now in flight at once, so "the blob in
  flight" is not a single thing and a per-blob bar would show eight uploads
  overwriting each other. `Event::Blob` / `Event::Bytes` are re-documented to
  say so. A blob the far side already holds is counted done the moment it is
  announced, so the bar does not stall further from the end with every skip.
- The **staging preflight** sizes as many finished layers as the transport
  uploads at once beside the state pass — largest first, capped by how many
  the stele has — because a finished layer holds its staging file until its
  own round trip lands. `StagingPeak::concurrent_other_bytes` replaces
  `largest_other_bytes`.

**Operator surface.** `dolos snapshot publish` and `dolos snapshot backfill`
gain `--concurrency N` and `--verify-carried`. `registry::Tuning` carries them
through `registry::open`; every non-publishing caller passes
`Tuning::default()`.

Out of scope, per the plan: layer geometry, and any registry Worker change.

## Verification

- `cargo test --workspace --all-targets --all-features` — green with the three
  crates CI excludes (`dolos-minibf`, `dolos-minikupo`, `dolos-trp`); see
  Findings.
- `cargo +nightly fmt --all --check` — clean.
- The `#[ignore]`d registry suites for `stelae` and `dolos-snapshot`, run
  locally against **`registry:2`** and **`zot:v2.1.20`** — the two ends of
  CI's matrix — and in CI against all three: **40 tests per registry, all
  passing**, including the crash-mid-publish resume tests and the
  killed-registry-restore test.

New tests:

- `concurrency_changes_the_cost_of_a_publish_and_not_the_stele` — the same
  records published serially and eight-way give the same inscription, the same
  identity, the same manifest bytes and the same transfer counters.
- `a_publish_dropped_mid_flight_leaves_the_previous_stele_standing` — a
  transport dropped with layers in flight and no seal leaves `latest` pointing
  at a stele that restores.
- `a_verified_carry_refuses_a_blob_the_repository_does_not_hold` — with
  `verify_adopted`, the refusal still lands before anything is tagged
  (replaces the old `adopt_layer`-time refusal).
- `a_layer_is_carried_forward_on_the_manifest_that_names_it` — the default
  path spends no round trip.
- `a_deferred_upload_that_fails_fails_every_seal` — **not** `#[ignore]`d: it
  points the transport at a closed port, so the deferral and the sticky
  refusal are covered under plain `cargo test` rather than only where a
  container can be spawned.

## What this PR cannot verify

The plan's done criterion is read off the live mainnet run's curve — the
+3 s/epoch slope flattening and absolute publish time dropping severalfold
after the driver image bump. That is a post-merge, ops-side measurement
against `~/stelae-backfill/mainnet-backfill.log`; nothing in this repository
can stand in for it. The plan's own contingency applies if it regresses: pin
the previous image tag, one epoch of rework.

Registry Worker behaviour under eight concurrent upload sessions is likewise
unmeasured here — the container fixture is not the deployment. Per the plan,
a Worker-side limit surfacing (429s, `cpu_ms` pressure) is a founder
escalation, not something to work around here; the modest default and
`--concurrency 1` are the levers if it does.

## Findings (pre-existing, not caused by this change)

- `main`'s CI is red on **Test (windows-latest)**, since #1259. Untouched here.
- `cargo test --workspace --all-targets --all-features` without CI's
  exclusions fails 314 tests in `dolos-minibf` (`epoch_value.rs:236`, accounts
  routes). Reproduces on `main`; CI's all-features job excludes that crate, so
  it does not gate anything.
- `crates/snapshot/tests/publish.rs` carried four
  `clippy::doc_lazy_continuation` warnings under the repo's pinned 1.93 (CI's
  clippy job installs 1.91 and does not see them). The doc comment they were
  in is rewritten here, so they are gone incidentally.

https://claude.ai/code/session_0162zpAPq7FPY6FNHFbPZi1h

## Summary by CodeRabbit

- **New Features**
  - Added configurable publish concurrency, including serial publishing.
  - Added verification for carried-forward layers.
  - Added snapshot backfill options for concurrency and layer verification.
  - Added configurable retry attempts and progress notifications for transient registry failures.

- **Bug Fixes**
  - Improved staging estimates for concurrent uploads.
  - Progress tracking now accurately aggregates transferred and retried bytes.
  - Added clearer errors when a layer cannot be published.
  - Improved cancellation handling during retry backoff.

### DIFF
--- a/crates/snapshot/src/registry.rs
+++ b/crates/snapshot/src/registry.rs
@@ -137,6 +137,7 @@ use std::{
     cell::{Cell, RefCell},
     collections::BTreeMap,
     io::Write as _,
+    num::NonZeroUsize,
     path::{Path, PathBuf},
 };
 
@@ -163,7 +164,7 @@ use stelae::{
 /// [`Auth`] rides along for the same reason: a host resolving its own
 /// credentials should not have to name the protocol crate to say what it
 /// resolved them to.
-pub use stelae::oci::{Auth, Registry, Repository, SCHEME};
+pub use stelae::oci::{Auth, Registry, Repository, DEFAULT_CONCURRENCY, SCHEME};
 
 use crate::{
     export::{self, history_for, same_network, Plan, Predecessor, Standing},
@@ -290,14 +291,19 @@ where
 /// `scratch_dir` is where layers are staged, in both directions. The transport
 /// creates it when the first layer needs it, so it need not exist yet.
 ///
+/// `tuning` is the publish path's concurrency and the one check an operator may
+/// want back; [`Tuning::default`] is what every caller that is not publishing
+/// wants.
+///
 /// **Never call any of this from inside an async context.** The transport owns
-/// a current-thread runtime and enters it with `block_on`; `stelae::oci`'s
-/// module documentation states the rule and the reason.
+/// a runtime and enters it with `block_on`; `stelae::oci`'s module
+/// documentation states the rule and the reason.
 pub fn open(
     repository: &Repository,
     insecure: bool,
     auth: Auth,
     scratch_dir: PathBuf,
+    tuning: Tuning,
 ) -> Result<Registry, Error> {
     Ok(Registry::open(
         repository,
@@ -305,8 +311,30 @@ pub fn open(
             insecure,
             scratch_dir: Some(scratch_dir),
             auth,
+            concurrency: tuning.concurrency.map_or(DEFAULT_CONCURRENCY, Into::into),
+            verify_adopted: tuning.verify_adopted,
         },
     )?)
+}
+
+/// What an operator may set about *how* a publish moves, as against where it
+/// goes.
+///
+/// Separated from [`open`]'s other arguments because it is the only one of them
+/// a caller can leave alone: a repository, a credential and a staging directory
+/// are facts a publish cannot be run without, and these two are a default and
+/// an escape hatch. A restore or an inspection passes [`Tuning::default`] and
+/// means it.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct Tuning {
+    /// How many layer round trips run at once; `None` is
+    /// [`DEFAULT_CONCURRENCY`].
+    pub concurrency: Option<NonZeroUsize>,
+
+    /// Re-prove that the registry still holds each blob carried forward out of
+    /// the predecessor's manifest. See [`stelae::oci::Options::verify_adopted`]
+    /// for why this is off.
+    pub verify_adopted: bool,
 }
 
 /// Where a publish is going, and what the host running it knows about itself.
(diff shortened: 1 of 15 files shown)


scarmuega:
**code-qa pass** — `evaluate-pr-comments` + `normalize-comments` applied to this branch.

**Review comments: 1 collected, 1 fixed, 0 escalated, 0 rejected.**

- `src/bin/dolos/feedback.rs:262-272` (CodeRabbit, outside diff range) — **fixed** in d700846b. Confirmed against the code: the block did describe the removed per-blob bar and argued against the cumulative one the same arm now implements, including a `set_position(0)` that no longer exists. Removed; the newer block stands alone.

**Comment sweep (PR scope):** one further fix in the same commit — `staging_peak` restated the saturating-arithmetic rationale that this PR had just lifted onto `StagingPeak::bytes`, where the type now states it once. No separator comments were added by this PR. The commit's diff is comment-only.

**Checks re-run at `d700846b`:** `cargo +nightly fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --workspace --all-targets`, and `cargo test --workspace --all-targets --all-features --exclude dolos-minibf --exclude dolos-minikupo --exclude dolos-trp` — all pass. The four new registry-backed tests were verified to actually run (not skip) against `registry:2`, `registry:3` and `zot` in the Registry workflow.

**Findings, no skill fixes them (for the reviewer, not this PR's scope):** pre-existing separator comments — which `normalize-comments` rules out — survive in files this PR touches but outside its diff: `crates/snapshot/tests/publish.rs` (4), `crates/stelae/tests/oci.rs` (3 plus `// --- up ---` / `// --- down ---` at 1412/1472), and `crates/stelae/Cargo.toml:39`. A codebase-scope sweep is the right vehicle, not this branch.

scarmuega:
**Field data on transient registry 5xx, and a retry ask** (mainnet backfill, night of 2026-08-23→24): the registry Worker returned sporadic `500 INTERNAL_ERROR` on `POST /v2/cardano/mainnet/blobs/uploads/` ~8 times over 11 h. On the current driver each one is fatal — the Job restarts, restores from `latest`, and re-replays the epoch. That was ~8 epochs of rework overnight (~7% of wall clock), all from a failure class that lasts milliseconds.

Reading this PR's semantics: a layer whose round trip fails is permanent for the transport (its staging went with the round trip), and recovery is *another publish*. Two asks, either of which absorbs the failure class:

1. **Retry inside the round trip**: bounded retry with backoff on transient 5xx/timeouts *before* the layer is declared lost — i.e., while the staged bytes are still in hand. Composes naturally with the worker pool (each worker retries its own layer), and keeps `LayerNotWritten` for genuinely non-transient refusals.
2. **Honor "recovery is another publish" in-process**: `backfill.rs` currently `?`s the publish error up, which turns a transient 500 into pod-restart → registry restore → full epoch re-replay — the most expensive recovery for the cheapest failure. The local stores are intact and consistent at that point; the loop could back off and re-run the publish step in place, exactly the recovery the transport's own docs prescribe.

(The registry-side root cause of the 500s is being chased separately — but the driver shouldn't need a minute-by-minute-healthy registry to make progress.)

PR #1263, opened 2026-08-25
fix(minibf): report null active stake for the preprod epoch gap

### LINKED ISSUE #1264: fix(minibf): report null active stake for the preprod epoch gap
Synthetic Kanban issue for:
- #1263

michalrus:
Resolves #1264.

## Problem

The `/epochs/{number}` route reported a genesis-derived active stake for preprod epochs 13-28. The reference reports `null` for these epochs, so the values did not match.

Preprod has a gap in its early stake snapshot. The reference reports active stake for epochs 6-12, `null` for epochs 13-28, then real values from epoch 29. Dolos writes stake logs continuously from epoch 6, so it fills the gap with the genesis sum.

## Implementation

This fix resets the active stake to `null` for preprod epochs 13-28. Preview and mainnet have no such gap and do not change.

## Testing

Found by:
- <https://github.com/blockfrost/blockfrost-tests/blob/229a9a1b8cea33446856fd57ac6b9d4ba4240481/src/fixtures/preprod/epochs/number-next.ts#>,
- <https://github.com/blockfrost/blockfrost-tests/blob/229a9a1b8cea33446856fd57ac6b9d4ba4240481/src/fixtures/preprod/epochs/number-previous.ts>.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected active-stake reporting for preprod epochs 13–28 to show no value when appropriate.
  - Preserved existing active-stake values for other epochs and networks.
  - Added coverage for the affected epoch range, its boundaries, and unaffected networks.
  - Improved consistency when active-stake data is unavailable from supported sources.

### DIFF
--- a/crates/minibf/src/hacks.rs
+++ b/crates/minibf/src/hacks.rs
@@ -415,3 +415,41 @@ pub fn maybe_set_genesis_previous_block<D: Domain>(domain: &Facade<D>, block: &m
         block.previous_block = Some(genesis_hash.to_string());
     }
 }
+
+/// This module lists the epochs where the reference indexer reports a `null`
+/// active stake, but Dolos computes a genesis-derived value. The reference
+/// indexer is db-sync, which the Blockfrost API uses.
+///
+/// The minibf epoch mapper sums the per-pool `StakeLog`s to report the active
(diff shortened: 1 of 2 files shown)


vladimirvolek (approved):
LGTM!

PR #1316, opened 2026-09-09
fix(api): share block metadata resolution through core

scarmuega:
UTxO listing created a fresh `AsyncQueryFacade` for each metadata lookup, giving each blocking decode its own semaphore and defeating the per-request limit. UTxO models, account-address assets, and minikupo matches now use a request-scoped `dolos_core::async_query::BlockMetaResolver` that owns one query facade and shares its blocking limiter.

Both API crates use their existing core dependency. The resolver lives in the existing `async_query` module alongside `AsyncQueryFacade` and `BlockRefMeta`; it adds no module or production dependencies and uses core's existing `futures-util`. API crates remain independent of each other.

The resolver deduplicates each batch, memoizes hits and misses in a bounded 4096-entry store, and exposes a fetch counter. Owned batch results remain complete above the memo capacity; eviction can only cause a later batch to re-fetch. Minikupo filters outputs before resolution and retains its existing missing-metadata error behavior. Ordering, pagination, and request acceptance remain unchanged; request-size policy is deferred.

Four core integration tests cover hit/miss deduplication, eviction and oversized batches, archive errors, and limiter sharing. Fixtures write synthetic blocks directly to the archive, without API test helpers or epoch-sensitive ledger imports.

Validation: formatting, the workspace all-target/all-feature build, default workspace all-target tests, and the feature-enabled workspace suite (with the repository's service-crate exclusions) all pass. Workspace clippy completes with seven existing warnings in unchanged cardano/snapshot tests. Tests ran outside the macOS sandbox to avoid its known system-configuration panic. `cargo tree -p dolos-minikupo --edges normal` confirms no minibf dependency.


crates/minikupo/Cargo.toml
```diff
@@ -24,6 +24,7 @@ axum = { version = "0.8.4" }
 
 dolos-core = { path = "../core" }
 dolos-cardano = { path = "../cardano" }
+dolos-minibf = { path = "../minibf" }
```
scarmuega: this is an architectural regression. API layers such as kupo and minibf can't depend on each other. Shared mechanism should be hoisted to existing upstream crates.

PR #1349, opened 2026-09-15
fix(minibf): answer unmatched paths with Blockfrost's global 400

### LINKED ISSUE #1323: minibf: unmatched paths return empty `404`, Blockfrost returns global `400` "Invalid path"
## Problem

MiniBF answers every unmatched path with an empty-body `404`. Blockfrost answers every unmatched path with a global `400`:

```json
{"error":"Bad Request","message":"Invalid path. Please check https://docs.blockfrost.io/","status_code":400}
```

This applies to any URL that matches no route: unknown endpoints, missing path segments, and empty path segments.

## Measured (2026-09-10, live `cardano-mainnet.blockfrost.io/api/v0`)

| URL | Blockfrost | dolos minibf |
| --- | --- | --- |
| `/nonexistent` | `400` Invalid path | `404`, empty body |
| `/foo/bar/baz` | `400` Invalid path | `404`, empty body |
| `/txs` | `400` Invalid path | `404`, empty body |
| `/blocks/latest/nope` | `400` Invalid path | `404`, empty body |
| `/blocks/epoch/2/slot` | `400` Invalid path | `404`, empty body |
| `/blocks/slot/` | `400` | `404`, empty body |

## Ground truth in the reference backends

- **ryo** (production): the app-level not-found handler returns the `400` body above. It comes from `@blockfrost/blockfrost-utils` (`lib/fastify.js`, `notFoundHandler`) and is registered in `src/app.ts` via `setNotFoundHandler`. So this is application behavior, not an external gateway.

## Suggested fix

Add a router-wide `.fallback()` in `crates/minibf/src/lib.rs` that returns the ryo `400` body. One handler covers all ~90 routes. Decide whether to also mirror ryo's trailing-slash and empty-segment handling, or leave those to the fallback.

## Notes

- Surfaced by Copilot's review of #1322; the divergence predates that PR and applies to every minibf route.
- Today's `404` responses also carry an empty body. Blockfrost error responses always carry a JSON body. The fallback fix covers unmatched paths; matched-route error bodies are a separate concern.

## Effort

> **Low** — easy mapping or reshaping of existing data, or a plain tx-body decode.

This is a reshape of the response.

arbelonson-source:
> **Disclosure:** this PR was written and opened by an autonomous coding agent (Claude Fable 5.1, via Claude Code) working on Arbel's account (`arbelonson-source`), taking the `agent-friendly` label on #1323 as the invitation. Please review it as agent output; if you'd rather not have agent contributions here, say so and I'll close it.

Closes #1323.

## Summary

Every path that matched no route got an empty-body `404`. Blockfrost (ryo's `notFoundHandler`) answers those with a global `400`:

```json
{"error":"Bad Request","message":"Invalid path. Please check https://docs.blockfrost.io/","status_code":400}
```

This adds a router-wide `.fallback()` that returns that body. The text lives in a new `Error::InvalidPath` variant, next to the other Blockfrost error bodies in `error.rs`, and the handler is a two-line `routes::invalid_path`.

## Semantics

| URL | Before | Now |
| --- | --- | --- |
| `/nonexistent`, `/foo/bar/baz`, `/txs`, `/blocks/latest/nope`, `/blocks/epoch/2/slot` | `404`, empty body | `400` Invalid path |
| `/blocks/slot/` | `404`, empty body | `400` Invalid path (the trailing slash is trimmed first, then nothing matches) |
| A matched route with a missing item, e.g. `/blocks/{unknown hash}` | `404` | `404`, unchanged |

On the issue's open question: trailing-slash and empty-segment handling is left to the existing `NormalizePathLayer` plus this fallback. `/blocks/latest/` still resolves to `/blocks/latest` as before, and whatever does not resolve gets the `400`. A wrong method on a matched path (`405`) is not touched.

## Testing

- Two tests in `crates/minibf/src/lib.rs`: the six URLs from the issue's table all get the `400` with the exact body, and `/blocks/{missing hash}` still gets its `404`, so the fallback does not swallow handler not-founds.
- Reverted check: with the `.fallback()` line removed, the new test fails with `unexpected status 404 Not Found for /nonexistent` and an empty body, which is the behavior the issue reports.
- `cargo test -p dolos-minibf`: 461 pass. `cargo +nightly-2026-08-27 fmt --all -- --check` and `cargo clippy -p dolos-minibf --all-targets --all-features -- -D warnings` are clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Unmatched paths now return a structured HTTP 400 “Bad Request” response with the message “Invalid path.” instead of an empty 404 response.
  - Requests using unsupported HTTP methods now receive the same consistent invalid-path response.
  - Trailing slashes are normalized before routing, improving route matching consistency.


crates/minibf/src/lib.rs
```diff
@@ -674,6 +674,7 @@ where
             "/governance/proposals/{gov_action_id}/parameters",
             get(routes::governance::proposal_parameters_by_gov_action::<D>),
         )
+        .fallback(routes::invalid_path)
```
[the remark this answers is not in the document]
michalrus: Hmmmm, this seems rather arbitrary. E.g. what about `"/blocks/epoch"`? Should `"epoch"` try to be parsed as `{hash_or_number}`, but `"slot"` not? 🤨

@slowbackspace, WDYT?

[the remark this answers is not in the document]
arbelonson-source: > (posted by Arbel's coding agent, Claude Fable 5.1 via Claude Code, same as the PR)

@michalrus agreed, and I'd rather not special-case `slot`. Copilot's premise also doesn't hold here: `NormalizePathLayer` is attached with `Router::layer`, which runs *after* routing (per the axum docs, to normalize before matching the layer has to wrap the `Router`; `build_router_with_facade` returns a plain `Router` with the layer inside it). So a trailing slash is never trimmed before route matching, in the server or in the tests, which go through the same `build_router_with_facade`.

Measured on this branch with a throwaway test through `TestApp`:

| path | status | message |
|---|---|---|
| `/blocks/slot/` | 400 | Invalid path (fallback) |
| `/blocks/slot` | 400 | Missing or malformed block hash. (`/blocks/{hash_or_number}`) |
| `/blocks/epoch` | 400 | Missing or malformed block hash. |
| `/blocks/epoch/` | 400 | Invalid path (fallback) |
| `/blocks/latest/` | 400 | Invalid path (fallback) |
| `/blocks/latest` | 200 | |

`/blocks/slot` and `/blocks/epoch` behave the same, and exactly as before this PR; the fallback only changes what unmatched paths get. `/blocks/slot/` is in the test list because it is one of the paths measured against Blockfrost in #1323, where it gets the global 400.

The `/blocks/latest/` row is a pre-existing quirk this exposes: since the normalizer never trims before matching, any trailing slash misses its route (an empty 404 before, the Blockfrost-style 400 now). Whether that should become a 200 depends on what Blockfrost does with a trailing slash, which I can't measure without a project token. Happy to file it separately, or fold a fix in here if you'd like, but it seemed like its own decision.

Unrelated: the `cargo deny` job fails the same way on the last three `main` runs (`override toolchain '1.93-x86_64-unknown-linux-musl' is not installed` inside the cargo-deny-action container).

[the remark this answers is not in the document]
michalrus: Yeah, I think so, too, but let's wait for @slowbackspace.

Re: `cargo-deny`, it's fixed in:
- #1344

slowbackspace, line 678: 🔴 **Trailing slashes are not trimmed before routing.** `NormalizePathLayer` is added with `Router::layer` at line 710. Axum runs that layer after route matching, so the trim never affects which route matches. Before this PR, `/blocks/latest/` returned an empty `404`. With this fallback it returns `400` "Invalid path" for a valid endpoint. Blockfrost returns `200`. The description claim that `/blocks/latest/` still resolves is not what happens.

Fix: wrap the router instead of layering it, `NormalizePathLayer::trim_trailing_slash().layer(app)`, and return that service. Commit c985eabe did exactly this before #860 moved it to `Router::layer`. The `ServiceExt::<Request>::into_make_service(app)` call at line 732 is the leftover of that wrapping and already fits the wrapped type.

slowbackspace, line 678: 🟡 **Wrong method still gets an empty `405`.** Blockfrost answers `POST /blocks/latest` with the same "Invalid path" `400`. One line closes it:

```rust
.method_not_allowed_fallback(routes::invalid_path)
```

The description marks this as out of scope, so treat it as optional.

crates/minibf/src/lib.rs
```diff
@@ -736,3 +737,57 @@ where
...
+
+    /// The paths measured against Blockfrost on the issue: unknown
+    /// endpoints, missing segments and an empty trailing segment.
+    const UNMATCHED_PATHS: &[&str] = &[
+        "/nonexistent",
+        "/foo/bar/baz",
+        "/txs",
+        "/blocks/latest/nope",
+        "/blocks/epoch/2/slot",
+        "/blocks/slot/",
```
slowbackspace: 🟡 **This row encodes the layer-ordering bug.** Once trimming runs before routing, `/blocks/slot/` matches `/blocks/{hash_or_number}` and returns `400` "Missing or malformed block hash." That is also what Blockfrost returns. The issue table lists only a bare `400` for this row, not the "Invalid path" body.

Drop the row, or assert the block-hash body for it. Also add a positive check that `/blocks/latest/` returns `200`, so the trailing-slash behavior has a regression guard.

crates/minibf/src/lib.rs
```diff
@@ -736,3 +737,57 @@ where
...
+                    "error": "Bad Request",
+                    "message": "Invalid path. Please check https://docs.blockfrost.io/",
+                }),
+                "unexpected body for {path}"
+            );
+        }
+    }
+
+    #[tokio::test]
+    async fn matched_route_still_reports_a_missing_component_as_not_found() {
```
slowbackspace: 🟢 **Duplicate test.** `blocks_by_hash_or_number_not_found` in `routes/blocks.rs` already asserts the same `404` for the same 64-`f` hash. This copy also pays a second full synthetic chain build. Delete it.

slowbackspace (changes requested):
Thanks, the fallback itself is small and correct. One real problem hides behind it, plus a few test points.

Severity legend: 🔴 must fix before merge · 🟡 should fix · 🟢 nit / optional

slowbackspace (commented):
One more point on the error text.

crates/minibf/src/error.rs
```diff
@@ -194,6 +197,15 @@ impl IntoResponse for Error {
                 )),
             )
                 .into_response(),
+            Error::InvalidPath => (
+                StatusCode::BAD_REQUEST,
+                Json(ErrorBody::new(
+                    400,
+                    "Bad Request",
+                    "Invalid path. Please check https://docs.blockfrost.io/",
```
slowbackspace: 🟡 **Drop the Blockfrost docs link.** Dolos should not send users to `https://docs.blockfrost.io/`. No other error message in this crate does. Keep the message as `Invalid path.` and adjust the test body to match.

crates/minibf/src/lib.rs
arbelonson-source: You're right, and thanks — I had this wrong in my earlier note. Fixed in 9293b5a5: the router is now wrapped with `NormalizePathLayer::trim_trailing_slash().layer(app)` instead of `Router::layer`, so trimming runs before route matching. `/blocks/latest/` returns 200 (matching Blockfrost), `/blocks/slot/` normalizes into `/blocks/{hash_or_number}` → 400 "Missing or malformed block hash.", and only genuinely unmatched paths reach the fallback. `ServiceExt::<Request>::into_make_service` at the serve call already fits the wrapped type, as you noted. Added `trailing_slash_is_trimmed_before_routing` as a regression guard.

— (posted by Arbel's coding agent, Claude via Claude Code, same as the PR)

arbelonson-source: Included it in 9293b5a5 — `.method_not_allowed_fallback(routes::invalid_path)`, so `POST /blocks/latest` now returns the same 400 "Invalid path." body as an unknown route, with a `wrong_method_answers_with_blockfrost_invalid_path` test. Happy to drop it if you'd rather keep this PR to the original scope.

crates/minibf/src/lib.rs
arbelonson-source: Done — dropped `/blocks/slot/` from `UNMATCHED_PATHS` (it's a matched route now), and the new `trailing_slash_is_trimmed_before_routing` test asserts both the `/blocks/latest/` 200 and the `/blocks/slot/` block-hash 400.

crates/minibf/src/lib.rs
arbelonson-source: Deleted in 9293b5a5 — `blocks_by_hash_or_number_not_found` in routes/blocks.rs already asserts the 404 for the same 64-`f` hash, so the copy (and its extra chain build) is gone.

crates/minibf/src/error.rs
arbelonson-source: Dropped — the message is now just `Invalid path.` in error.rs, and the fallback test body matches.

arbelonson-source:
Pushed 9293b5a5 addressing the review — thanks @slowbackspace, the layer-ordering catch was the real one and I had it wrong in my earlier note.

- 🔴 **Trailing slashes** — wrapped the router (`NormalizePathLayer::trim_trailing_slash().layer(app)`) so trimming runs before routing. `/blocks/latest/` → 200 (Blockfrost parity), `/blocks/slot/` → the `/blocks/{hash_or_number}` route's 400 "Missing or malformed block hash.", and only genuinely unmatched paths reach the fallback.
- 🟡 **Wrong method** — added `.method_not_allowed_fallback(routes::invalid_path)`; `POST /blocks/latest` now returns the invalid-path 400. (Optional per scope — easy to drop if you'd prefer.)
- 🟡 **Test row** — dropped `/blocks/slot/` from the unmatched list; added a `trailing_slash_is_trimmed_before_routing` regression test asserting the `/blocks/latest/` 200 and `/blocks/slot/` 400, plus a wrong-method test.
- 🟢 **Duplicate test** — removed; `blocks_by_hash_or_number_not_found` already covers it.
- 🟡 **Docs link** — message is now just `Invalid path.`

All 461 minibf tests pass; `cargo clippy -p dolos-minibf --all-targets -- -D warnings` and `cargo +nightly-2026-08-27 fmt --all -- --check` are clean. Ready for another look when you have a moment.

crates/minibf/src/lib.rs
```diff
@@ -343,7 +348,7 @@ impl<D: Domain> Facade<D> {
 
 pub struct Driver;
 
-pub fn build_router<D>(cfg: MinibfConfig, domain: D) -> Router
+pub fn build_router<D>(cfg: MinibfConfig, domain: D) -> NormalizePath<Router>
```
michalrus: Hmmm, I'd rather you didn't change the public `Router` type, `NormalizePath<_>` is an implementation detail. It is possible to apply `NormalizePath`, and keep the `Router` outer type, I've seen it somewhere, hmmm.

Ah, found it:

```rust
    let inner = NormalizePathLayer::trim_trailing_slash().layer(inner);
    let app = Router::new()
        .fallback_service(inner);
```

michalrus (changes requested):

crates/minibf/src/lib.rs
arbelonson-source: Good call — done in a04ea0b4. `build_router` / `build_router_with_facade` are back to returning `Router`; the normalizer is applied with `NormalizePathLayer::trim_trailing_slash().layer(app)` and nested behind an outer `Router` via `fallback_service`, exactly your snippet, so `NormalizePath<_>` stays an implementation detail. Behaviour unchanged — 462 minibf tests pass, clippy `-D warnings` and nightly fmt clean.

crates/minibf/src/lib.rs
slowbackspace: drop it. 405 sounds good to me in this case even if it doesn't match blockfrost.

arbelonson-source: Dropped in 9716b985 — removed `method_not_allowed_fallback` and its test, so a wrong method is back to the default 405. Agreed, keeping it simple.

arbelonson-source:
Disclosure: I'm an autonomous coding agent (Claude Opus 5.5, via Claude Code) on Arbel's account.

@michalrus — your `Router` point is in a04ea0b4 (`build_router` returns `Router` again, normaliser nested via `fallback_service`), and @slowbackspace's 405 note is in 9716b985. I re-checked against today's `main` after #1354 landed: it merges without conflicts, and on the merged tree `cargo test -p dolos-minibf` is 489 passed / 0 failed with clippy `-D warnings` clean. Ready for another look whenever you have time.

crates/minibf/src/error.rs
```diff
@@ -194,6 +197,15 @@ impl IntoResponse for Error {
                 )),
             )
                 .into_response(),
+            Error::InvalidPath => (
+                StatusCode::BAD_REQUEST,
+                Json(ErrorBody::new(
+                    400,
+                    "Bad Request",
+                    "Invalid path.",
```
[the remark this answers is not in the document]
michalrus: Wrong, it's hard to say that the issue "requires" the link. @slowbackspace rejected precisely that in https://github.com/txpipe/dolos/pull/1349#discussion_r4024202883.

crates/minibf/src/lib.rs
```diff
@@ -706,7 +708,13 @@ where
...
             CorsLayer::new()
         });
-    app.layer(NormalizePathLayer::trim_trailing_slash())
+    // Added via `Router::layer`, NormalizePathLayer runs *after* route matching
+    // and never trims before routing (`/blocks/latest/` would miss its route).
+    // Wrap the router with it and nest that behind an outer `Router` via
+    // `fallback_service`, so trimming happens before routing while the public
+    // return type stays `Router` (the wrapped type is an implementation detail).
+    let normalized = NormalizePathLayer::trim_trailing_slash().layer(app);
+    Router::new().fallback_service(normalized)
```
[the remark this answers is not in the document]
michalrus: Wrong. Issue #1323 lists `/blocks/slot/` as a bare `400`, not as "Invalid path". Without the trailing slash, this path matches `/blocks/{hash_or_number}`, which returns a `400` with "Missing or malformed block hash." @slowbackspace asked for exactly this behavior, and for this test, in https://github.com/txpipe/dolos/pull/1349#discussion_r4024199303.

michalrus (approved):

PR #1351, opened 2026-09-15
fix: pad full era history for gRPC and Ogmios consumers

### LINKED ISSUE #1331: gRPC/UTxO-RPC `ReadEraSummary` doesn't return the full historical era table (unlike minibf's `/network/eras`)
## Summary

As of 1.6.0, `ReadEraSummary` (`src/serve/grpc/v1alpha/query.rs`) only reflects whatever era-boundary records the backing node has itself locally processed since it started tracking chain state — it doesn't reconstruct the full historical era table (Byron..Conway) the way a full node, Blockfrost's `/network/eras`, or Dolos's own minibf `/network/eras` route does.

`crates/minibf/src/routes/network.rs` already pads this out with the full table — the gRPC/UTxO-RPC and Ogmios (`src/serve/o7s_unix/statequery.rs`) interfaces could reuse that same logic so all three interfaces agree.

## Impact

A client whose era interpreter expects a complete, fixed-shape era table can't parse a partial response at all — not just less accurate, but unusable, forcing a hardcoded per-network fallback client-side instead.

## Suggested fix

Reuse minibf's era-padding logic in the gRPC and Ogmios interfaces.

adrian1-dot:
## Summary

Fixes #1331. `ReadEraSummary` (gRPC v1alpha/v1beta) and the Ogmios n2c `GetInterpreter` query only returned eras the node had actually recorded as on-chain state:
- Byron is never recorded as tracked state on mainnet/preprod (genesis starts post-Byron).
- Preview's genesis skips Shelley, Allegra, and Mary entirely, jumping straight to Alonzo.

minibf's Blockfrost-compatible `/network/eras` route already worked around this with its own hardcoded padding logic. This PR extracts that into a shared `dolos_cardano::pad_era_history` helper and wires it into both gRPC query handlers and the Ogmios state-query server, so all three surfaces (gRPC, Ogmios, minibf) return the same complete Byron-through-tip era table. minibf's route is refactored to call the shared helper instead of duplicating the hardcoded logic.

## Testing

- New unit tests in `crates/cardano/src/eras.rs` cover mainnet Byron padding/chaining, Preview's triple-era skip, intra-era protocol-bump dedup, unsupported network-magic rejection, and the zero-recorded-eras edge case.
- `cargo clippy --workspace --all-targets --all-features` — clean on the changed files.
- Manually verified against a live preprod node: queried `ReadEraSummary` via gRPC (both v1alpha and v1beta) with `grpcurl`, and cross-checked every era boundary against Koios's `epoch_params.protocol_major` transitions and Blockfrost's `/network/eras`. All boundaries (epochs 4/5/6/7/12/163) and parameters (`epoch_length`, `slot_length`, `safe_zone`) matched exactly across all three independent sources.

## Summary by CodeRabbit

* **New Features**
  * Era history views now reconstruct complete historical timelines, including eras not explicitly recorded on-chain.
  * Historical results extend through the current chain tip across gRPC and state-query interfaces.
  * Preview networks now show appropriate skipped-era placeholders and historical boundaries.
  * Network era metadata includes improved start/end times, slots, and safe-zone values.

* **Bug Fixes**
  * Intra-era protocol changes are no longer shown as duplicate eras.
  * Unsupported network configurations now return clear errors instead of producing invalid history data.


adrian1-dot:
**Note for reviewers: duplicated era-name grouping**

While fixing the CI failures this PR turned up (see commit history), I found that the `protocol → era name` grouping (`5,6 → alonzo`, `7,8 → babbage`, `9,10 → conway`, etc.) is now hardcoded independently in **four** places:

- `src/serve/grpc/v1alpha/query.rs::protocol_to_era_name`
- `src/serve/grpc/v1beta/query.rs::protocol_to_era_name`
- `src/serve/o7s_unix/statequery.rs` (inline match)
- `crates/cardano/src/eras.rs::era_group` (added by this PR, for `pad_era_history`'s dedup)

The first three predate this PR; this PR adds a fourth copy rather than consolidating. This duplication is directly related to one of the bugs this PR fixes — `pad_era_history`'s old dedup logic assumed a fixed set of "opening" protocol literals, which drifted out of sync with reality (Preview's genesis records Alonzo starting at protocol 6, not 5) and silently dropped the era.

Happy to follow up with a small refactor that moves this grouping into one canonical function in `dolos_cardano` that all four call sites import, so the padding logic and the display-naming logic can't drift apart again — but wanted to flag it here rather than expand this PR's scope unilaterally. Let me know if you'd like that as a follow-up PR, part of this one, or if you'd rather leave it as-is.

*(Separately: `cargo deny`'s failure on this PR is a pre-existing yanked-crate warning on `spin` via `flume`→`fjall`, unrelated to this change — visible on other open PRs too.)*

Issue #1296, opened 2026-09-02
minibf: Refactor and cleanup after all endpoints are implemented


vladimirvolek:
The endpoints were implemented by several people working in parallel, one
endpoint at a time, and the layout still reflects that. This is a tidy-up pass:
give each endpoint set its own folder, pull the parts they share into one place
instead of repeating them, and keep tests next to the code they cover.

Nothing about the API changes - same paths, same responses, same status codes.

One set per PR.

- [ ] accounts
- [ ] addresses
- [ ] assets
- [ ] blocks
- [ ] epochs
- [ ] governance
- [ ] metadata
- [ ] network
- [ ] pools
- [ ] scripts
- [ ] txs
- [ ] utils
- [ ] utxos
- [ ] genesis
- [ ] health
- [ ] metrics

PR #1354, opened 2026-09-16
refactor(minibf): blocks

### LINKED ISSUE #1355: refactor(minibf): blocks
Synthetic Kanban issue for:
- #1354

vladimirvolek:
Resolves #1355.

We have 100% of the blocks routes implemented!

`blocks.rs` becomes `routes/blocks/`:

- every endpoint gets its own file, with its tests at the bottom
- `mod.rs` keeps what the endpoints share (id parsing, block loading, tip lookup, tx paging) and re-exports the handlers, so `lib.rs` stays as it is
- duplicated code is pulled into one helper instead of being copied around

Code that is not HTTP moves out of `routes/`:

- `BlockModelBuilder` → `src/mapping/blocks.rs`
- synthetic genesis block → `src/hacks/genesis_block.rs`, the three per-network copies collapsed into one function

Blockfrost fixtures for `blocks` pass the same before and after.

+ a lot more tests

part of https://github.com/txpipe/dolos/issues/1296

## Summary by CodeRabbit

- **New Features**
  - Added block lookup by hash, number, slot, or epoch and slot.
  - Added latest-block retrieval and navigation to previous or next blocks.
  - Added paginated transaction hashes, transaction CBOR, and touched-address endpoints.
  - Added synthetic genesis-block support for supported networks.
  - Block responses include chain context, neighboring blocks, transaction details, fees, outputs, and confirmations.
- **Bug Fixes**
  - Improved validation and error handling for malformed, missing, and unavailable block data.
  - Preserved pagination, ordering, and genesis attribution across block endpoints.


slowbackspace (commented):
Sound refactor. The split is faithful and I would approve it after the `/blocks/latest` fix.

I diffed the old `routes/blocks.rs`, `mapping.rs` and `hacks.rs` against all 17 new files. The handler bodies for `previous`, `next`, `addresses`, `txs`, `txs_cbor` and `by_epoch_slot`, plus `BlockModelBuilder`, are faithful copies. Status codes match everywhere: `404` on a missing block, `503` on an empty archive, `400` on a bad id. The genesis hack applies in exactly the same five handlers as before. Collapsing the three per-network genesis copies into `genesis_for_domain` is correct — hashes, times and `next_block` values all match the originals.

I found nothing wrong in the pagination, ordering, genesis-in-range or touched-address logic.

**Severity legend:** 🔴 blocking · 🟡 should fix · 🟢 nice to have

crates/minibf/src/routes/blocks/latest.rs
```diff
@@ -0,0 +1,47 @@
...
+
+use crate::Facade;
+
+use super::{single_block_content, tip_block};
+
+pub async fn latest<D>(State(domain): State<Facade<D>>) -> Result<Json<BlockContent>, StatusCode>
+where
+    D: Domain + Clone + Send + Sync + 'static,
+{
+    let tip = tip_block(&domain)?;
```
slowbackspace, line 13: 🔴 `/blocks/latest` now reads the tip twice. This line calls `tip_block`, then `single_block_content` calls `tip_block` again at `routes/blocks/mod.rs:158`. The old handler passed one tip as both arguments: `build_block_model(&domain, &tip, &tip, &chain)`.

Two problems:

- `get_tip` returns a full `BlockBody` from flatfiles. This doubles the read cost on the hottest blocks endpoint.
- The two reads are not atomic. If a block lands between them, the response describes block N but reports `confirmations: 1`. Blockfrost always reports `0` for `/blocks/latest`, and the old code did too.

Suggested fix: give `single_block_content` an optional tip parameter, or add a `single_block_content_with_tip` that `latest` calls with the tip it already holds.

crates/minibf/src/hacks/mod.rs
```diff
@@ -57,15 +58,6 @@ struct GenesisTxModel<'a> {
...
-pub fn genesis_hash_for_domain<D: Domain>(domain: &Facade<D>) -> Option<&'static str> {
-    match domain.genesis().shelley.network_magic {
-        Some(1) => Some(GENESIS_HASH_PREPROD),
-        Some(2) => Some(GENESIS_HASH_PREVIEW),
-        Some(764824073) => Some(GENESIS_HASH_MAINNET),
-        _ => None,
-    }
-}
-
 pub fn genesis_block_metadata_for_domain<D: Domain>(
```
slowbackspace: 🟡 The genesis time table now exists twice. This function keeps its own `network_magic -> (hash, time)` match, and `hacks/genesis_block.rs:22` adds `PREPROD`/`PREVIEW`/`MAINNET` consts with the same times.

The hash values are shared consts, so only `time` can drift. But a change to one copy would desync `/blocks/0` from `/txs/{genesis-tx}`. Point this function at the new consts instead.

crates/minibf/src/routes/blocks/by_slot.rs
```diff
@@ -0,0 +1,31 @@
...
+    Json,
+};
+use blockfrost_openapi::models::block_content::BlockContent;
+use dolos_core::{ArchiveStore as _, Domain};
+
+use crate::Facade;
+
+use super::single_block_content;
+
+pub async fn by_slot<D>(
```
slowbackspace, line 13: 🟢 This is the only extracted handler with no test module. Every sibling got the 4-test matrix; `next` got 8 and `previous` got 7. The 404, 500 and happy paths of `by_slot` stay uncovered.

slowbackspace (approved):
lgtm

PR #1361, opened 2026-09-17
chore: bump pallas to 1.3.0 and render script JSON as text

scarmuega:
## Summary

Adopts pallas 1.3.0, which carries the native script and Plutus data stack-safety series, and switches the one Dolos code path that still built a recursive value tree.

Two commits:

1. **chore: bump pallas to 1.3.0.** Workspace pin `1.1.1` to `1.3.0`, lock file updated for the twelve pallas crates. What this brings in:
   - txpipe/pallas#802: native script CBOR decoding no longer overflows the stack. Preprod block 5183974 carries a 5,383-deep script.
   - txpipe/pallas#803: `ScriptNOfK`'s threshold is the ledger's signed `int64`. Preprod tx `b1db2a41…` in block 5184855 has a threshold of `-1` and stopped sync permanently.
   - txpipe/pallas#805: the u5c native script mapper is iterative, so follow-tip and fetch-block no longer crash on that script.
   - txpipe/pallas#806: `to_json` builds iteratively and `to_json_string` writes text without a value tree; shared `TreeNode` traversal in pallas-codec.
   - txpipe/pallas#807: `PlutusData` decoding no longer overflows the stack. A datum costs one CBOR byte per level, so a single transaction can carry roughly 16,000 levels against a previous ceiling near 3,250 on a 2 MiB thread.
   
   The `ScriptNOfK` field type change is breaking upstream but Dolos does not touch that field. `NativeScript` now has a custom `Drop`; Dolos only hashes it, never destructures it.

2. **fix(minibf): render script JSON without building a value tree.** `/scripts/{hash}/json` now calls `to_json_string` and wraps the text in the response body directly. At chain-realistic depth a `serde_json::Value` cannot be serialized or dropped on a 2 MiB worker stack even when built iteratively, so this is the only shape that closes the endpoint crash. Response shape is unchanged: `{"json": …}` with `null` for Plutus scripts, and the existing route tests still parse it into the Blockfrost `ScriptJson` model.

## Verification

- `cargo check --workspace --all-targets` clean against 1.3.0, no new warnings.
- Workspace library tests pass, including all 28 script route tests.
- Clippy and rustfmt clean on the touched crate.

## Not covered here

Serve-path visitors over `PlutusData` (the u5c datum and metadata mappers, datum JSON) and `Metadatum` decoding still recurse in pallas; those are the next pallas PRs and will arrive through a later bump.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved script JSON responses to avoid failures when rendering deeply nested native scripts.
  * Responses now include the correct `application/json` content type.
  * Plutus scripts consistently return a JSON response with a `null` value when no native-script representation is available.

* **Tests**
  * Added coverage for response headers and Plutus script JSON behavior.
  * Expanded synthetic test fixtures to include Plutus V2 scripts.


scarmuega:
@coderabbitai resolve

PR #1368, opened 2026-09-18
feat(minibf): add `/governance/dreps/{drep_id}/votes`

### LINKED ISSUE #1091: minibf: add `/governance/dreps/{drep_id}/votes`
Implement the Blockfrost `/governance/dreps/<drep>/votes` endpoint in the `minibf` crate.

### Effort

~L — High effort (needs a data-layer change)~

M — Medium effort (`minibf` + archive) — after 480e5deb284381e799de90dda4c182e99b332185 (`v1.7.0-alpha.1`+13)

### Approach

~No vote store~

`drep_votes` reverse scan over proposals ns; `tx_hash`/`cert_index` via block-at-slot

### Why now

0.0071% of observed traffic.

michalrus:
Resolves #1091.

## Implementation

- Read DRep vote history from proposal state.
- Resolve transaction data from archived blocks.
- Add pagination tests and synthetic vote fixtures.

## Testing

The `blockfrost-tests` suite tests this endpoint in these files:

- https://github.com/blockfrost/blockfrost-tests/blob/0f3f3ea268983a6cf2e6f7fd24c5082fc3a91e0e/src/fixtures/preprod/governance/dreps/drep-votes.ts
- https://github.com/blockfrost/blockfrost-tests/blob/0f3f3ea268983a6cf2e6f7fd24c5082fc3a91e0e/src/fixtures/preview/governance/dreps/drep-votes.ts

## Summary by CodeRabbit

- **New Features**
  - Added an endpoint to retrieve votes cast by a specified DRep, with pagination and scan-limit support.
  - Vote results include the associated proposal and voting transaction details.

- **Bug Fixes**
  - Updated scan-limit error messaging to accurately describe request item limits.

- **Documentation**
  - Documented the new DRep votes endpoint in the API coverage reference.


crates/minibf/src/routes/governance/mod.rs
```diff
@@ -1491,6 +1759,189 @@ mod tests {
...
+    async fn governance_drep_votes_orders_and_paginates() {
+        let app = drep_votes_app();
+        let drep = &app.vectors().drep_id;
+        let ascending = get_drep_votes(&app, drep, "").await;
+        let descending = get_drep_votes(&app, drep, "?order=desc").await;
+
+        assert_eq!(
+            descending.iter().map(|row| &row.tx_hash).collect_vec(),
+            ascending.iter().rev().map(|row| &row.tx_hash).collect_vec()
+        );
```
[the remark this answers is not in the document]
michalrus: Fixed in 764cede9b913af0774396f2e32097bee9efe8793.

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -689,6 +712,248 @@ where
...
+
+    let page = page_slot_groups(
+        rows,
+        |row| row.slot,
+        pagination,
+        |slot, group| settle_casts(domain, &voter, slot, group),
+    )?;
+
+    page.into_iter()
+        .filter_map(|row| row.cast.map(|cast| (cast, row)))
```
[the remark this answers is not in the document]
michalrus: Fixed in 0d244816ef8f205773f017083e76c593dc657a0b.

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -689,6 +712,298 @@ where
...
+    if descending {
+        groups.reverse();
+    }
+
+    let from = pagination.from();
+    let mut skipped = 0;
+    let mut out = Vec::with_capacity(pagination.count);
+
+    for mut group in groups {
+        settle_casts(domain, voter, group[0].slot, &mut group)?;
```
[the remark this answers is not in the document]
michalrus: Fixed in a353fb4c3cdedff6719004e201af151203c9f11d.

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -689,6 +712,307 @@ where
...
+            continue;
+        };
+
+        for (slot, vote) in history {
+            rows.push(VoteRow {
+                slot: *slot,
+                proposal_tx: state.tx,
+                proposal_idx: state.idx,
+                vote: vote.clone(),
+                cast: None,
```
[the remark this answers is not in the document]
michalrus: Fixed in 81dd024289982f9b8009874aaccf6547ba55bb7a.

crates/minibf/src/error.rs
```diff
@@ -172,7 +171,7 @@ impl IntoResponse for Error {
                 Json(ErrorBody::new(
                     400,
                     "Bad Request",
-                    "The requested page needs more archive blocks than this node is willing to scan, reduce page number or count.",
+                    "The request has more items than this node permits.",
```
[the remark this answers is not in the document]
michalrus: Fixed in a6e0690c567f300ceb24a91918a9fe602fa61213.

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -689,6 +712,312 @@ where
...
+
+    for entry in entities {
+        let (_, state) = entry.map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+
+        let Some(history) = state.drep_votes.get(voter) else {
+            continue;
+        };
+
+        for (slot, vote) in history {
+            if rows.len() >= budget {
```
slowbackspace: **Severity: 🟡 medium**

This budget counts the DRep's whole vote history, not the request. A DRep with more than `max_scan_items` rows (default 3000) gets a 400 on every request, including `?count=1&page=1`. The error text says "Reduce the page number or the count", but no page or count can help.

The budget also protects nothing. The namespace walk runs to the end before the check can matter, and `read_page` for `/governance/proposals` walks the same namespace without a budget. The archive work is already capped by `enforce_max_scan_limit`.

Preview holds ~1545 proposals today, so a test DRep that re-votes crosses 3000 rows. Suggest dropping this check and the two tests that pin it (`rejects_history_over_scan_budget`, `accepts_history_at_scan_budget`).

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -689,6 +712,312 @@ where
...
+    let mut skipped = 0;
+    let mut out = Vec::with_capacity(pagination.count);
+
+    for (scanned, mut group) in groups.into_iter().enumerate() {
+        if scanned == budget {
+            return Err(Error::ScanBudgetExceeded);
+        }
+
+        settle_casts(domain, voter, group[0].slot, &mut group)?;
+        group.retain(|row| row.cast.is_some());
```
slowbackspace: **Severity: 🟡 medium**

Because unresolved rows are removed before counting (0d244816), this loop settles every slot group before the page window. `?count=100&page=30` on a full-archive node decodes up to ~2900 blocks per request. `page_slot_groups` in the same file avoids exactly this: it skips groups before the window and only reads blocks for rows on the page.

The extra cost buys stable offsets on a `max_history` node, where votes in pruned blocks vanish anyway. The alternative is to settle only the groups inside the window and drop unresolved rows there. Pages on a pruned node come back short, which is what `max_history` already means for the other history endpoints, and offsets stay aligned with Blockfrost. That path can reuse `page_slot_groups` with a `settle` closure that calls `settle_casts` and retains resolved rows, so the generic gets its second caller back.

Either way, the choice deserves a sentence in the PR description. The long-term fix is to record `(tx_hash, cert_index)` in the vote history at roll time (`visit_tx` has the tx), but that is a schema revision and belongs in a follow-up.

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -689,6 +712,312 @@ where
...
+    if descending {
+        groups.reverse();
+    }
+
+    let from = pagination.from();
+    let mut skipped = 0;
+    let mut out = Vec::with_capacity(pagination.count);
+
+    for (scanned, mut group) in groups.into_iter().enumerate() {
+        if scanned == budget {
```
slowbackspace, line 888: **Severity: 🟢 low**

This branch is unreachable. `vote_page` guarantees `rows.len() <= budget`, so `groups.len() <= budget` and `scanned` never reaches `budget`. The doc comment above says the budget limits resolved slot groups, which is no longer true. The test that covered this branch (`governance_drep_votes_stops_at_scan_budget`) was removed in 81dd0242; `rejects_deep_page` only exercises `enforce_max_scan_limit`.

Remove the `budget` parameter and this check, or keep them and drop the rows budget in `vote_page` so this one does the work again (with its test restored).

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -689,6 +712,312 @@ where
...
+    // Proposal history stores rows in cast order. Each removal takes the first
+    // matching cast in block order.
+    let mut settled: Vec<((usize, u32), VoteRow)> = rows
+        .drain(..)
+        .map(|row| {
+            match casts
+                .get_mut(&(row.proposal_tx, row.proposal_idx))
+                .and_then(VecDeque::pop_front)
+            {
+                Some((position, tx, cert_index)) => (
```
slowbackspace, line 838: **Severity: 🟡 medium**

No test exercises the pairing this `VecDeque` exists for: one DRep voting on the same proposal in two transactions of one block. In `drep_votes_config`, block 1 votes on three distinct proposals and block 2 casts one vote, so every deque holds one entry. If the pairing regressed (a plain `HashMap` overwriting the first cast, or a sort that loses history order), both rows would report the later `tx_hash` and every current test would still pass.

Add a block where the key DRep votes on action `(0,0,0)` in tx 0 and again in tx 1, and assert the two rows carry `tx_hashes[0]` then `tx_hashes[1]` with `cert_index` 0 each.

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -689,6 +712,312 @@ where
...
+                    (position, cert_index),
+                    VoteRow {
+                        cast: Some((tx, cert_index)),
+                        ..row
+                    },
+                ),
+                None => ((usize::MAX, u32::MAX), row),
```
slowbackspace: **Severity: 🟢 low**

This arm also fires when the archive has the block but the block has no matching cast for the voter. That is a state/archive mismatch, not a pruned block, and it is silently dropped by the `retain` in `page_resolved_votes`. A `tracing::warn!` with slot, proposal and voter here would make the divergence visible.

slowbackspace (commented):
Reviewed against `blockfrost-backend-ryo` (`dreps_drep_id_votes.sql`) and the pinned `blockfrost-tests` fixtures. `blockfrost-mimicry` has no vote endpoints, so ryo is the only reference. Clippy is clean and the 71 governance tests pass on `a6e0690c` in an isolated worktree.

What checks out: the listing order matches ryo (`voting_procedure.id` = tx order in the block, then `GovActionId` map order). `cert_index` matches db-sync, which numbers each voter's ballot from 0 in `Map.toList` order; pallas decodes the ballot as `BTreeMap<GovActionId, _>` with the same `Ord`. Unknown and special DReps return `200 []` like ryo.

Legend: 🔴 high = fix before merge (correctness, perf) · 🟡 medium = design or parity risk · 🟢 low = cleanup, tests, nits

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -689,6 +712,312 @@ where
...
+    vote: Vote,
+
+    /// The transaction that contains the vote and its index in that
+    /// transaction. These values stay empty until `settle_casts` processes the
+    /// row group.
+    cast: Option<(Hash<32>, u32)>,
+}
+
+/// This function converts a DRep ID to the proposal vote-map key.
+fn drep_credential(drep_bytes: &[u8]) -> Result<StakeCredential, StatusCode> {
```
slowbackspace: **Severity: 🟢 low**

`parse_drep_id` builds the 29-byte id from `(prefix, payload)`, and this function splits it again and re-derives the variant from the nibble. `drep_voter` then repeats the mapping that `ProposalState::vote_history_mut` already encodes. Returning the `StakeCredential` from `parse_drep_id` (or adding an inverse of `drep_to_entity_key` in `dolos-cardano`) would keep the id-to-credential rule in one place.

crates/minibf/src/routes/governance/mod.rs
michalrus: Correct, the budget counted the whole history and not the request, so a DRep past `max_scan_items` got a 400 on every page. Removed the budget and its two tests, `max_scan_items` now limits only the page depth. Fixed in 119a9b794279189f45ce4909ed2cd6a8e3343993.

crates/minibf/src/routes/governance/mod.rs
michalrus: Took the alternative you proposed. `vote_page` calls `page_slot_groups` with a settle closure, and filters the unresolved rows after paging, so offsets stay the same as in Blockfrost and a pruned node returns a short page. Fixed in 119a9b794279189f45ce4909ed2cd6a8e3343993.

crates/minibf/src/routes/governance/mod.rs
michalrus: Correct, the branch was unreachable. It went away with `page_resolved_votes`. Fixed in 119a9b794279189f45ce4909ed2cd6a8e3343993.

crates/minibf/src/routes/governance/mod.rs
michalrus: Added `governance_drep_votes_pairs_repeated_votes_within_one_block`. It casts two votes on one proposal from two transactions of one block, and makes sure that row 0 names tx 0 and row 1 names tx 1. Fixed in 119a9b794279189f45ce4909ed2cd6a8e3343993.

crates/minibf/src/routes/governance/mod.rs
michalrus: Added a `tracing::warn!` with the slot, the voter, and the proposal id. Fixed in 119a9b794279189f45ce4909ed2cd6a8e3343993.

crates/minibf/src/routes/governance/mod.rs
michalrus: Added `drep_from_entity_key` to `dolos-cardano` as the inverse of `drep_to_entity_key`, and `ProposalState::vote_history` as the read side of `vote_history_mut`. The endpoint now uses both, so the header rule and the voter-key rule each live in one place. Fixed in 119a9b794279189f45ce4909ed2cd6a8e3343993.

crates/minibf/src/routes/governance/mod.rs
michalrus: The PR description got a new `## Pagination` section.

michalrus:
[Copilot's "Archive reconstruction"](https://github.com/txpipe/dolos/pull/1368#pullrequestreview-5292368129) above is:

https://github.com/txpipe/dolos/blob/119a9b794279189f45ce4909ed2cd6a8e3343993/crates/minibf/src/routes/governance/mod.rs#L775-L791

vladimirvolek (approved):

scarmuega (approved):

michalrus:
I'll resolve conflicts in a minute ⏳

PR #1369, opened 2026-09-19
feat(minibf): add the constitutional committee vote endpoints

### LINKED ISSUE #1108: minibf: add `/governance/committee/{cc_id}/votes`
Implement the Blockfrost `/governance/committee/<comittee>/votes` endpoint in the `minibf` crate.

### Effort

~L — High effort (needs a data-layer change)~

M — Medium effort (`minibf` + archive) — after 480e5deb284381e799de90dda4c182e99b332185 (`v1.7.0-alpha.1`+13)

### Approach

~No committee state / vote store~

As #1110; cold→hot resolves through `committee_auths` history

### Why now

0.0007% of observed traffic.

### LINKED ISSUE #1110: minibf: add `/governance/committee/votes`
Implement the Blockfrost `/governance/committee/votes` endpoint in the `minibf` crate.

### Effort

~L — High effort (needs a data-layer change)~

M — Medium effort (`minibf` + archive) — after 480e5deb284381e799de90dda4c182e99b332185 (`v1.7.0-alpha.1`+13)

### Approach

~No committee state / vote store~

`cc_votes` scan; anchor + height/time from the same block decode

### Why now

0.0006% of observed traffic.

michalrus:
Resolves #1108.

Resolves #1110.

## Implementation

- Add `/governance/committee/votes` and `/governance/committee/{cc_id}/votes`. Both routes share one scan of the committee vote histories in the proposal namespace. The scan keeps a bounded frontier of complete slot groups. It then reads each block from the archive for the transaction hash, the anchor, the block height, and the order inside the block.

- `/governance/committee/{cc_id}/votes` accepts a CIP-129 hot or cold credential. Votes carry the hot credential only. A cold credential therefore resolves through the authorization history in `committee_auths`.

- ⚠️ [**resync!**] Do not add the `CommitteeGc` delta at the epoch boundary. This delta erased the authorization history of every past committee member. A cold credential of a past member therefore resolved to nothing.

- Ratification does not change. `active_committee_size` and `committee_tally` iterate the current members and read the m      ap by key. The retained entries stay unread. The map grows by one entry for each past committee member.

## Testing

The synthetic test chain now carries a vote anchor.

The `blockfrost-tests` suite tests this endpoint in these files:
- https://github.com/blockfrost/blockfrost-tests/blob/0f3f3ea268983a6cf2e6f7fd24c5082fc3a91e0e/src/fixtures/preprod/governance/committee/votes.ts
- https://github.com/blockfrost/blockfrost-tests/blob/0f3f3ea268983a6cf2e6f7fd24c5082fc3a91e0e/src/fixtures/preview/governance/committee/votes.ts
- https://github.com/blockfrost/blockfrost-tests/blob/0f3f3ea268983a6cf2e6f7fd24c5082fc3a91e0e/src/fixtures/preprod/governance/committee/cc-id-votes.ts
- https://github.com/blockfrost/blockfrost-tests/blob/0f3f3ea268983a6cf2e6f7fd24c5082fc3a91e0e/src/fixtures/preview/governance/committee/cc-id-votes.ts


crates/cardano/src/ewrap/loading.rs
```diff
@@ -1500,10 +1500,6 @@ impl BoundaryWork {
             self.add_delta(crate::GovDormancyTick::new());
```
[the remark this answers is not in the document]
michalrus: Okay, fair, fixed in 24a95939eab32fe97fd410545539fcc5ef2e02d3.

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -476,6 +477,459 @@ where
...
+
+        if page.len() == pagination.count || !truncated {
+            break page;
+        }
+
+        if group_limit == budget {
+            return Err(Error::ScanBudgetExceeded);
+        }
+
+        group_limit = (group_limit * 2).min(budget);
```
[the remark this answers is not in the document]
michalrus: Okay, fixed in f4876b5cf9453e9f0deecf355c3a3e7d0e71781c.

crates/cardano/src/model/gov.rs
```diff
@@ -1014,13 +1043,39 @@ impl dolos_core::EntityDelta for CommitteeGc {
...
+                    archived.ends_with(history),
+                    "committee authorization archive must end with the GC history"
+                );
+                archived.truncate(archived.len() - history.len());
+                archived.is_empty()
+            };
+
+            if remove_archive_entry {
+                state.committee_auth_archive.remove(cold);
+            }
```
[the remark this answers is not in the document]
michalrus: The panic cannot occur: a `CommitteeGc` delta never reaches the WAL, because the boundary work unit keeps the default, empty `commit_wal`, and only `RollWorkUnit` writes a `LogValue` (`roll/batch.rs:275`). Thus, the rollback in `core/sync.rs:125` reads block deltas only, and `_ => false` would hide a real invariant violation instead of a legacy case, so I kept the assertion and recorded this reason in the code.

You are correct about the stale text: the note beside `SCHEMA_REVS` now gives the reason for `gov` at revision 2, but `lib.rs:192` is the shard count and not the revision, so it stays at 1. Fixed in f66b3b6e9e185af986f6db065f597bbd5f3b82d5.

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -476,6 +477,464 @@ where
...
+            continue;
+        };
+
+        for (voter, history) in &state.cc_votes {
+            if voters.is_some_and(|voters| !voters.contains(voter)) {
+                continue;
+            }
+
+            for (slot, vote) in history {
+                if scanned == budget {
```
slowbackspace, line 782: **Severity: 🔴 high**

For `/governance/committee/votes` this counter charges every committee vote in chain history against `max_scan_items`, whatever the page. Once the chain holds more than 3000 committee votes (the default), the endpoint returns 400 for `?count=1&page=1` and for every other request, permanently. Mainnet has 7 members voting on ~160 actions so far; preview and preprod test committees vote far more often. The only remedy is to raise `max_scan_items`, which also loosens the page cap on every other endpoint.

The counter does not protect anything the walk does not already pay. `read_page` for `/governance/proposals` decodes the same namespace without a budget. The frontier bounds memory to `group_limit` groups, and archive reads are bounded by `group_limit <= to() <= max_scan_items`. Rows inside one slot are bounded by the block size.

Suggest dropping the row budget and the two tests that pin it (`rejects_history_over_scan_budget`, `counts_rows_inside_one_slot`), and keeping the `group_limit == budget` guard as the loop terminator. Same point as https://github.com/txpipe/dolos/pull/1368#discussion_r4073169319, but here the trigger is certain.

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -476,6 +477,464 @@ where
...
+    if descending {
+        groups.reverse();
+    }
+
+    let mut skipped = 0;
+    let mut out = Vec::with_capacity(pagination.count);
+
+    for mut group in groups {
+        settle_committee_casts(domain, group[0].slot, &mut group)?;
+        group.retain(|row| row.cast.is_some());
```
slowbackspace, line 731: **Severity: 🟡 medium**

Because unresolved rows are dropped before the page offset is counted, this loop settles every group from the frontier edge up to the page. The frontier holds `to()` groups, so `?count=100&page=30` decodes up to 3000 blocks per request on a full-archive node. If rows counted toward the offset before settling, only the groups that overlap `[from, to)` would need a block read (at most `count + 1`), and pruned nodes would return short pages instead of shifted ones.

Same trade-off as https://github.com/txpipe/dolos/pull/1368#discussion_r4073169328; whichever way #1368 lands, the two endpoints should agree, and the choice deserves a sentence in the PR description.

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -476,6 +477,464 @@ where
...
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+enum CommitteeCredentialRole {
+    Hot,
+    Cold,
+}
+
+/// This function parses a CIP-129 constitutional-committee credential. The
+/// header byte gives the role and the credential type. If the header role and
+/// the Bech32 prefix do not agree, the credential is not valid.
+fn parse_committee_id(id: &str) -> Result<(CommitteeCredentialRole, StakeCredential), Error> {
```
slowbackspace, line 489: **Severity: 🟡 medium**

This is the inverse of `bech32_committee_hot` / `bech32_committee_cold` in `crates/minibf/src/mapping.rs`, but no test round-trips the pair. The route tests parse only a hot script id and a cold key id. The hot key URL in `bad_request` fails on `page=0` before the parser runs, and no test parses a cold script id. A regression in the `0x2` branch for `cc_hot` or the `0x3` branch for `cc_cold` would pass the suite.

Please add a unit test that encodes all four (role, kind) combinations with the encoders and asserts the parse returns the same role and credential, plus a 28-byte and a 30-byte payload and a low nibble of `0x0`/`0x1` as rejections. Moving the parser next to `bech32_committee` would also let both sides share the HRP constants and the nibble values instead of repeating the literals here.

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -2122,6 +2601,458 @@ mod tests {
...
+
+    /// A budget that holds every row starts with a frontier that is too
+    /// small for the page. The retry must increase the frontier. It must
+    /// then return the full page in the correct order.
+    #[tokio::test]
+    async fn governance_committee_votes_grows_a_truncated_frontier() {
+        let app = TestApp::new_with_scan_limit(committee_votes_across_slots_config(), 4);
+        let blocks = app.vectors().blocks.clone();
+
+        // The first frontier holds two groups, because `count` is 2. The
```
slowbackspace, line 2983: **Severity: 🟢 low**

This test never truncates. `group_limit = to().min(budget).max(1)`, and `to()` is `count * page` = 4 = budget, so the first frontier holds all four vote slots and the loop exits on the first pass. The comment says the first frontier holds `count` groups, which is not what the code does.

With a full archive the growth path is unreachable: `to()` groups always yield at least `to()` rows, so the page is full and the loop breaks. Growth only happens through pruning, which `grows_past_a_pruned_group` and `returns_a_short_page_at_the_group_budget` already cover. Either fix the comment and keep this as a plain paging check, or drop it.

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -476,6 +477,464 @@ where
...
+    match cred {
+        StakeCredential::AddrKeyhash(hash) => Voter::ConstitutionalCommitteeKey(*hash),
+        StakeCredential::ScriptHash(hash) => Voter::ConstitutionalCommitteeScript(*hash),
+    }
+}
+
+/// This function finds the transaction, the anchor, and the block height for
+/// each vote in one block. It then sorts the rows by transaction position and
+/// ballot index.
+fn settle_committee_casts<D: Domain>(
```
slowbackspace, line 637: **Severity: 🟢 low**

`settle_committee_casts` and `resolve_committee_vote_page` repeat the block-matching and page-skipping logic of `settle_casts` and `page_resolved_votes` from the DRep path: decode the block, build a `casts` map with a `VecDeque` per target, `pop_front` in history order, sort by (tx position, index), drop rows without a cast, reverse for desc, skip `from`. The committee version is a superset (voter set, anchor, block height). Once #1368 settles, generalizing the DRep helpers over a voter filter and a row type would leave one copy to fix.

crates/minibf/src/routes/governance/mod.rs
```diff
@@ -476,6 +477,464 @@ where
...
+                continue;
+            }
+
+            for (slot, vote) in history {
+                if scanned == budget {
+                    return Err(Error::ScanBudgetExceeded);
+                }
+                scanned += 1;
+
+                frontier.push(CommitteeVoteRow {
```
slowbackspace, line 787: **Severity: 🟢 low**

Every row is built here, with a clone of `voter` and `vote`, before `push` decides whether the frontier accepts it. On the aggregate endpoint most rows fall outside the frontier. A `would_accept(slot)` check on the frontier before building the row avoids the work. Separately, each retry of the loop in `committee_vote_page` rescans the whole namespace and re-decodes the blocks that the previous pass already settled; carrying settled groups across retries would make the pruned-node path cheaper.

slowbackspace (commented):
Reviewed against `blockfrost-backend-ryo` (`committee_votes.sql`, `committee_cc_id_votes.sql`, `validateCommitteeCredentialId`) and the pinned `blockfrost-tests` fixtures. On `f66b3b6e` in an isolated worktree: clippy is clean for the files this PR touches, the 83 minibf governance tests, the 15 cardano committee tests and every snapshot suite pass.

What checks out: the listing order matches ryo's `voting_procedure.id` (tx order, then voter map order). Pallas orders `ConstitutionalCommitteeScript` before `ConstitutionalCommitteeKey`, which is the ledger's `Credential` order that db-sync inserts in, so the several-members-in-one-tx case matches. The cold-to-hot union over every recorded authorization matches ryo's `committee_registration` join. The 400 text is identical, unknown ids return `200 []`, and the `governance_type` names match ryo's snake_case mapping. The `gov` revision bump follows the schema-rev process (rev 1 golden kept, rev 2 pinned). Uppercase bech32 ids are fine: `CheckedHrpstring::hrp()` is guaranteed lowercase.

Legend: 🔴 high = fix before merge (correctness, perf) · 🟡 medium = design or parity risk · 🟢 low = cleanup, tests, nits

crates/minibf/src/routes/governance/mod.rs
michalrus: Fixed in 038338a20b12c7f7358340a3ef81296f95fbb529.

crates/minibf/src/routes/governance/mod.rs
michalrus: Fixed in 038338a20b12c7f7358340a3ef81296f95fbb529.

crates/minibf/src/routes/governance/mod.rs
michalrus: Fixed in 038338a20b12c7f7358340a3ef81296f95fbb529.

crates/minibf/src/routes/governance/mod.rs
michalrus: Fixed in 038338a20b12c7f7358340a3ef81296f95fbb529.

crates/minibf/src/routes/governance/mod.rs
michalrus: Fixed in 038338a20b12c7f7358340a3ef81296f95fbb529.

crates/minibf/src/routes/governance/mod.rs
michalrus: Fixed in 038338a20b12c7f7358340a3ef81296f95fbb529.

